PR #17, opened 2016-05-18
remaining cards for The Liberated Mind pack

JoelCFC25:
I'm not sure why the colons after `"limited"` are showing in red on cards 79-81...


Alsciende:
The test says:

```
Validating cards from The Liberated Mind...
/home/ubuntu/netrunner-cards-json/pack/tlm.json: File is not valid JSON.
Expecting ',' delimiter: line 42 column 9 (char 1141)
```

Alsciende:
The colons are red because the previous line lacks a comma at the end

JoelCFC25:
Ah, ok--fixed.

PR #45, opened 2016-06-03
zh: change colons in card names & paid abilities

stimhackcn:
from full-width to half-width so identity names and slogans display separately, like other languages. Paid abilities in accord with english version for possible further use.


Alsciende:
Can you create fewer Pull Requests please?

You can commit in your branch several times, and then when you're done, create a PR from that branch.

stimhackcn:
Havn't noticed it could be done that way. Sorry for the inconvenience! I'll figure it out from now.

PR #52, opened 2016-06-06
add templates for each of the card types

gereons:
I made some templates for each of the card types, maybe other editors will find them helpful.
(These aren't valid JSON by design).


Alsciende:
Very good idea, thanks!

PR #55, opened 2016-06-21
Update ts.kr.json

castigar:
Korean translation for "The Source" is done :)


Alsciende:
Please check the test after you submit the PR. The test failed because the JSON syntax is not correct.

castigar:
hmm, would you let me know how to check the test after i submit?
I'm not familiar with github :/

Alsciende:
Sure. On the very page of the Pull Request (this page), there's an automatic message that says if the checks passed or failed. That message disappears when the PR is merged, so you can't see it anymore. But you can still see that your commit is in error : there's a red cross in front of your commit (above my reply). If you click on the cross, you'll be redirected to a page that describes the checks and why they failed. The output is very verbose, but somewhere in those lines, you can see:

> home/ubuntu/netrunner-cards-json/translations/kr/pack/ts.kr.json: File is not valid JSON.
Expecting ',' delimiter: line 101 column 73 (char 3770)

That means there's an issue with your file at the line 101. The check wasn't able to parse your file because the file is not valid JSON. That's a serious problem, as it means NetrunnerDB won't be able to read and import the data.

As it turned out, the problem was that your typed `"\` instead of `\"`. I fixed the problem and commited the correct file.

Alsciende:
You can see how I fixed your file in this commit: https://github.com/zaroth/netrunner-cards-json/commit/d20cb087308b6bc417927425dfcfacc927340be7

castigar:
thanks buddy,
got it! lol

PR #56, opened 2016-06-22
Update oac.kr.json

castigar:
I can see "able to merge" message.
That should be the OK message for the update, right?  :)


Alsciende:
"Able to merge" means that there's no conflict between your files and my files. It's a good thing, but it doesn't mean your files pass the tests.

As it happens, your files don't pass the test. You should be able to see it, on this page.

Alsciende:
The good news is, at the end it says

> Found 1 formatting and 0 validation errors

If there are no validation errors, then you're good to go. Formatting errors don't really matter.

PR #58, opened 2016-06-23
Update oac.kr.json

castigar:


castigar:
I've found the error on line 227.
I should have not change the line :/

PR #59, opened 2016-06-23
Update val.kr.json

castigar:


castigar:
I have no idea what makes the formatting error pops up

Alsciende:
Please, create a Pull Request when you're finished editing a batch of files. That way, I only have 1 PR to check, instead of 5.

PR #64, opened 2016-06-27
Update kr.json uw/oh/uot/tlm

castigar:


castigar:
thanks for the merge, buddy.
finally, I've almost done every datapacks. :)
but is there any progress before the translation shows up on the site?
cuz it has not beed changed til now.

Alsciende:
I haven't deployed the new translations yet. In a few hours, I will.

PR #70, opened 2016-07-04
added some keywords

lpoulter:


lpoulter:
Why does the validate.py re-format every json file? Is this expected behavouir?

Alsciende:
It does so with the -f option

zaroth:
Leaving no spaces or newlines would also be valid JSON, technically, but the goal isn't to be only valid here. The goal is to be easily readable by humans :)

The formatting rules are AFAIK the default ones from python JSON formatter.

PR #71, opened 2016-07-04
changed li tags to new lines

lpoulter:
Looks like this should of been done with /n not \<li> slack displays \<li> as raw tags.


Alsciende:
I'm sorry, I can't merge this. <li> are perfect for this.

PR #73, opened 2016-07-06
zh 23 seconds translation

stimzh:


stimzh:
There's 1 formatting error but no validation error. Checked but couldn't sort it out.

Alsciende:
You're right, there was an issue with the importation of type, cycle, etc. I fixed it.

PR #74, opened 2016-07-06
Architect's "champs design" text in accord with 13's and 14's

stimzh:


Alsciende:
The more I think about it, the more I think it's better to put it in the flavor text. CBI Raid and Deuces Wild should be put in line with Architect, not the other way around. I'll try to put something up so that these special flavor text are displayed in bold instead of italic.

PR #76, opened 2016-07-12
Most Wanted List update

JoelCFC25:
According to today's release, the updated Most Wanted List goes into effect on 1 August.


Alsciende:
The new MWL is another MWL. I'll handle it.

PR #87, opened 2016-08-05
add missing flavor & trash cost, zh bm pack.

stimzh:
Please also compile es.zh.json and "neutral" in faction.zh.json.


Alsciende:
> Please also compile es.zh.json and "neutral" in faction.zh.json.

What do you mean? I've updated the zh locale. Can you tell me what is wrong?

Alsciende:
Ok I see. For example on Scarcity of Resources, the word "Neutral" is not translated.

stimzh:
Thank you. And yes, all corp and runner neutral cards are labeled "neutral" instead of "中立"(Chinese).

Alsciende:
That should be fixed now.

PR #90, opened 2016-08-05
Update asis.jp.json

LogicWolfJP:
One card for Japanese Champ Deck.


Alsciende:
Please create 1 PR for all your files ;)

PR #98, opened 2016-08-06
Update fc.jp.json

LogicWolfJP:
One card for Japanese Champ Deck


stimzh:
You can edit the whole file under your own fork and then create a single pull request containing all the changes, which would be less stressful for the operator.

PR #114, opened 2016-09-18
Add Net Mercur + Fix Ark's illustrator name

stimzh:


stimzh:
[http://www.peachhack.com/wp-content/uploads/2016/09/adn38-net-mercur.png](card pic)

PR #122, opened 2016-10-04
Adam flavor and rest card pics

stimzh:


Alsciende:
omg I'm so late on this one... thank you!

PR #130, opened 2016-11-04
fix card numbers of Hasty Relocation and Fumiko

JoelCFC25:
The positions/card numbers of Hasty Relocation and Fumiko Yamamori are mixed up.


lostie:
Good spot! 👍

Alsciende:
That's very annoying, as users now have Fumiko in their deck instead of Hasty Relocation, and vice-versa. Oh well...

JoelCFC25:
I only found it when I pulled data down to start coding Fumiko on jnet and ending up getting Hasty Relocation in my hand instead...

PR #143, opened 2016-12-09
Fix card definition for Jinteki ID AgInfusion

lostie:
As pointed by Joel, the IDs don't have the flavour text and the extra text is contained in the title itself.


lostie:
@JoelCFC25 👀

Alsciende:
Indeed! Thank you.

PR #151, opened 2017-01-14
Fix duplicate and misspelt artist names

ouroboros8:
As well as fixing some trivial misspellings, I've taken the liberty of homogenising a couple of names. Specifically, I've renamed all instances of `Zefanya Maega` to `Zefanya Langkan Maega`, and one instance of `Matthew Zeilinger` to `Matt Zeilinger`.

While the previous state of the data does accurately reflect the cards (e.g. Zefanya's middle name is used on Fast Track, but not on John Masanori), I feel that the ability to search by a single, consistent name across all cards is more useful than strict adherence to physical card text.

I chose the long-form of Zefanya's name on the basis that it was most recently used (in Honor and Profit, as opposed to previous short-form uses in the Spin Cycle), and chose Matt as opposed to Matthew Zeilinger because come on, who calls him Matthew Zeilinger.


Alsciende:
Very good job. Thank you!

PR #187, opened 2017-03-23
Fix Self-Adapting Code Wall and add Station One release date

dodgepong:


packs.json
```diff
@@ -266,7 +266,7 @@
     {
         "code": "so",
         "cycle_code": "red-sand",
-        "date_release": null,
+        "date_release": "2017-03-23",
```
Alsciende: I missed the announcement?

dodgepong: They might not have had an article this time (or the article hasn't been released), but the release date has been known for several weeks. Release dates are now included on the ["Upcoming" page](https://www.fantasyflightgames.com/en/upcoming/) when their status is changed to "Shipping Now".

PR #194, opened 2017-03-24
Update td.json

ANR-PeterPan:
Some more spoils from Four is flatline on BGG
Added Bloodletter https://boardgamegeek.com/camo/f08580609d7decdfa9d3629599c8853c9bb07381/687474703a2f2f692e696d6775722e636f6d2f41666a503052702e6a7067 and Hortum https://boardgamegeek.com/camo/e015d6eedca28e2a2651cbe3527b7f4089cc9334/687474703a2f2f692e696d6775722e636f6d2f4777426b4a31582e6a7067 from BGG. Moved K.P.Lynn to #52 because Upgrades are always last. Perhaps they added 2 cards after the interview like adding the link to Polyhisor. This would also explain why this "deluxe" has 57 different cards and not 55.

https://boardgamegeek.com/camo/d098e52e44d97092a32fe3574e2df1a29375efb6/68747470733a2f2f692e696d6775722e636f6d2f705137664369392e706e67
Syn Attack must be #4, but we only see a 2 cost Double Event.
Biometric Spoofing is #25-#27
I think Mason Bellamy is #40 or #38 if HB gets more than 2 Upgrades and still missing text. 
Added Hailstorm as #49 which is the only place for weyland ICE left, but Strength is hidden. I wrote "?", Plz correct if this is wrong.

#makeweylandgreatagain


Alsciende:
Changing card codes breaks a few things in the application. That should be avoided at all cost: only add cards to the db when you're absolutely sure about their codes. Better to wait a few weeks for official data.

Alsciende:
Also, '?' is not a valid strength. Again, it's better to wait for complete and official spoilers than updating the site with erroneous data.

PR #208, opened 2017-04-13
MWL 1.2

NoahTheDuke:
Updates to Most Wanted List 1.2. [Official announcement](https://www.fantasyflightgames.com/en/news/2017/4/12/enhanced-enforcement-tactics/).


dodgepong:
Is it worth tracking the "type" of MWL between 1.0/1 and 1.2, in terms of universal influence vs reducing max influence?

NoahTheDuke:
Probably not through this medium. Definitely necessary on netrunnerdb, but for this the quantity handles that.

lostie:
I had a quick look into the way the MWL is currently implemented on [NetrunnerDB](https://github.com/Alsciende/netrunnerdb) and I'm pretty sure there's some changes required in it so that this list works appropriately for deck building purposes. As it stands, all the cards in the new MWL will reduce max influence instead of using the universal influence.

I _**would suggest to hold this for now**_ until the new MWL behaviour is put in place in the [NetrunnerDB](https://github.com/Alsciende/netrunnerdb).

I like @dodgepong 's idea of having some way to define/identify the way the MWL needs to be handled, but I'll leave any decision to @Alsciende as he might have a much better idea how to put that in place.

Alsciende:
I'm going to add a property to the "mwl" object, describing the type of MWL.

I'm thinking "penalty_target". Versions 1.0 and 1.1 will have "penalty_target": "identity", while version 1.2 and up will have "penalty_target": "cards". What do you think?

Alsciende:
Scratch this, I'd prefer to store it as a boolean. What about "global_penalty", which is true in 1.0 and 1.1, and false in 1.2+.

yads:
I think the penalty_target approach is more extensible. Although maybe it may be more clear to use something like `penalty_type: <'subtract_global_influence'|'add_universal_influence'>`. That way, if a future MWL has a different way of doing things, you can add it to this enum field.

Alsciende:
> if a future MWL has a different way of doing things

Don't say that, you're going to jinx it!

NoahTheDuke:
Looks like it was jinxed anyways.

PR #225, opened 2017-07-24
Update and fix Japanese translations

dlrowb:
1) Fix keywords separations (added whitespaces)
2) Added Japanese translations (red-sand cycle)
3) Fix names so that there are just the ones that are actually sold


gereons:
please fix the validation errors in translations/jp/pack/eas.jp.json (i.e. run ./validate.py -f) so that we can merge this

dlrowb:
Now I remove whitespaces and pass validation :)

PR #249, opened 2017-08-31
deploy last fixes

Alsciende:


gereons:
Just out of curiosity, when are these changes expected to show up in the nrdb API? https://netrunnerdb.com/api/2.0/public/cards?_locale=de still has the empty _locale block for UVC, while the "Clan" subtype for Salvaged Vanadis Armory is already visible.

Alsciende:
The translations didn't update automatically. I updated it now. It should be live in ~10 minutes.

gereons:
Thanks a lot!

gereons:
Sorry, something seems to be off here - the _locale block is still empty.

Alsciende:
The translation is databased though https://netrunnerdb.com/de/card/13038

Alsciende:
I'll see what's going on

gereons:
Thanks, your efforts are very much appreciated, as always!

Alsciende:
Ok I busted the cache for that resource, it did the trick :)

PR #251, opened 2017-09-12
add Revised Core Set as as new cycle/pack "core2"

gereons:


JoelCFC25:
Are you sure you want this to be code 14? The Kitara cycle will definitely be coming out before Core 2.

gereons:
Even if it would come out later, since you can build it yourself right now *and* will be mandatory for worlds, it's better to put it in the slot before Kitara, IMO.

Alsciende:
I agree. Since we know its content and have its cards, it's virtually released already.

I was thinking we could jump to pack number 20 for Core 2.0, what do you think?

gereons:
Sure, if Kitara then gets 21. I think that would be necessary/helpful to make figuring out "cards up to xyz" not be a pain in the neck.

Alsciende:
Yes, with Kitara at 21.

gereons:
ok, I'll update the PR.

PR #255, opened 2017-09-17
added kitara cycle + sovereign sight previews

alexduryee:
Added information from https://www.fantasyflightgames.com/en/news/2017/9/11/sovereign-sight/.  It looks like Jinja City Grid and Friday Chip are from the next two packs (31 and 42, respectively), so they aren't included here.


gereons:
Please change the cycle number to 21 so that Kitara gets sorted after the revised core set.

gereons:
ss.json appears to contain a BOM, please remove. You can use the `validate.py` script in this repo to verify that all files are correctly formatted - it should report `Found 0 formatting and 0 validation errors`.

alexduryee:
Thanks!  Should be resolved.

gereons:
Thanks! There are a few more changes necessary before we can merge this, and those are the individual card codes. They need to start with "21" now as well, e.g. Lewi Guilherme needs to have "21005"

gereons:
Thank you!

Issue #257, opened 2017-09-20
New MWL - Ban / Restricted List


NoahTheDuke:
~~According to [4chan->reddit](https://www.reddit.com/r/Netrunner/comments/716fbg/spoilers_spoilerken_leaks_the_new_mwl_on_4chan/), the new MWL will be a drastic change, going from universal influence increases to a Ban / Restricted list the same as other LCG games (cannot include banned cards in your deck, can only include up to a playset of a single restricted card in your deck). Given the consistency of spoilers from the source (spoilerken has been correct all of the other times), might want to start bikeshedding about how to code this.~~

It's [official](https://images-cdn.fantasyflightgames.com/filer_public/af/3e/af3ed829-6e1b-4a40-a1b3-122fcd3f2b14/upcoming_adn_mwl_v20.pdf).

@Alsciende, how've you done it for Game of Thrones? Can you easily port it to Netrunner?

NoahTheDuke:
@Alsciende, looks like it's official.

PR #258, opened 2017-09-27
Updated to MWL 2.0

NoahTheDuke:
Modifies mwl.json to handle the new MWL 2.0 changes, as noted in #257.

I took the liberty of changing the "global_penalty" flag to be follow @yads's [comment](https://github.com/Alsciende/netrunner-cards-json/pull/208#issuecomment-293906340), set all of the Restricted cards to 3, and set all of the Removed cards to 0, under the idea that a number would denote how many of a given card can be included in a deck (leaving the Restricted exclusivity requirement to be implemented by each program, not by the data-set).

My changes are not and should not be final. I welcome all comments and suggestions on how to best modify this to fit our needs. Thankfully, the full list of cards has already been compiled, so all that would need changing is the penalty flag and the numbers.


yads:
I guess I jinxed it :) I think it looks good. I'm wondering if maybe keeping the global_influence flag on the old MWLs as a deprecated field makes sense, depending how hard it is to update the existing code.

yads:
Probably makes sense to keep it for backwards compatibility, but not have it in the 2.0 document.

yads:
It also might be better to make the card values contain the card's status, rather than quantity. For example, in order to find restricted cards, the code would currently need to infer information about its status. Perhaps something more along the lines of:

```
{
  "03038":  "restricted",
  "08061": "banned",
  ...
}
```

Or even if we wanted to make it even more extensible:

```
{
  "03038":  { "type": "restricted" },
  "08061": { "type": "banned" },
  ...
}
```

gereons:
I like the idea of marking types, but I'd rather not repeat it for every card. Why not invert the relationship, and have types with associated lists of cards?

```
{
  "restricted": [ "03038", ... ],
  "banned": [ "08061", ... ]
}
```

Alsciende:
The MWL could overload parts of the cards.

MWL version 1.0 and 1.1 overload a new `global_penalty` property of some cards.

MWL version 1.2 overloads a new `universal_faction_cost` property of some cards.

MWL version 2.0 overloads the `deck_limit` property of some cards, and a new `is_restricted` property of some other cards.

Applying a MWL is overloading the cards data with the mwl data: `cardData = Object.assign(baseCardData, mwlCardData)`

Whether or not a MWL is active, the properties are always considered. The application of the deck legality rules is separated from the activation of a MWL.

yads:
Ooh, I really like the idea of extending card properties based on the info in the MWL document. That way, like you said, you can look at the card data in whole, rather than looking in a bunch of different sources.

NoahTheDuke:
I'm not into bike-shedding; I have no preference on how the data should be presented. Just let me know how y'all want it done, and I'll whip it up!

Alsciende:
I'd like the data to look like

```
    {
        "cards": {
            "08061": {
                "deck_limit": 0
            },
            "03035": {
                "is_restricted": 1
            }
        },
        "code": "NAPD_MWL_2.0",
        "date_start": "2017-09-27",
        "name": "NAPD MWL 2.0"
    }
```

I'll store `cards` in a json array.

NoahTheDuke:
Want this back-ported to the previous MWLs?

Alsciende:
Yes. I don't think anyone is using that endpoint.

NoahTheDuke:
Updated.

jwarwick:
FWIW, jinteki.net uses that `mwl` endpoint. But we would need to make changes there to support new semantics anyway.

jwarwick:
Found two errors in the file as I was working on the jinteki.net import:
1. Muertos Gang Member (08068) was swapped for Gang Sign (08067).
2. Rumor Mill (11022) was missing.

Alsciende:
Thank you, @NoahTheDuke , it looks good.

@jwarwick would you prefer that I keep the /2.0/mwl endpoint untouched, and create a new endpoint for this new format?

jwarwick:
@Alsciende - I've already updated the Jinteki code, so go ahead and modify the endpoint. When my last pull request goes through we should be compliant with both the `rotated` flag and the MWL changes. Thanks for getting all that updated.

NoahTheDuke:
Updated.

PR #259, opened 2017-09-29
Discussion for New Data Format

NoahTheDuke:
After seeing @Alsciende's comment [here](https://github.com/Alsciende/netrunnerdb/issues/136#issuecomment-329402106) about how L5R data is set up and how he wishes NRDB was similar, I decided to try out switching the card data to the "new" format and see what everyone thinks.

I based the card and pack structures entirely off of how [fiveringsdb-data](https://github.com/Alsciende/fiveringsdb-data/) work. Because I wasn't sure if it would be good to switch from **code** identifiers (current) to name-based **id** identifiers (5RDB), I generated both and put them in the folders cards-by-code and cards-by-id, respectively.

I didn't include my generation script cuz it's a little hacky, but I could if y'all want it.

Obviously, this is not final and should not be how it actually looks. But I think for beginning the discussion, it's a good first step.


gereons:
My two cents: I would go the cards-by-id route, but unless I'm missing the point of having arrays everywhere I would not store each card in an array that always has to have exactly one element, but instead in a proper json object. So instead of saying

```
[
    {
        "code": "12345",
        ...
    } 
]
```

I'd prefer

```
{
    "card": {
        "code": "12345",
        ...
    } 
}
```

Alsciende:
Every data json file contains an array of objects, that's how I implemented in in fiveringsdb. That way, all files are parsed the same way and arrays can be merged or split at will. For example, all the cards data file can be merged to create an array with all the cards.

Alsciende:
One thing to consider with card-by-id is that it would break all existing card routes.

gereons:
Yeah, doing this would definitely mean bumping the API version and keeping 2.0 around for a reasonable deprecation period.

NoahTheDuke:
I personally think avoiding [link rot](https://en.wikipedia.org/wiki/Link_rot) by making as small a change as possible is smartest. This kind of change would only necessitate deprecating the Core 2.0 card codes, and then doing a one-time change of all card links in the decks, descriptions, and comments from Core 2.0 codes to the pack-agnostic codes.

Alsciende:
Using numeric codes would only break the Core2 cards. But using slug ids would break all the cards.

gereons:
But would you necessarily have to switch to slugs in the URLs just because the editing side does? The codes would be still present and also serve as a unique identifier for card.

Re: deprecation periods: I'm currently exploring options of how to de-couple Net Deck from the 3 API endpoints it needs for its basic data (cards, packs and cycles) to help reduce the time needed to keep old stuff around.

ANR-PeterPan:
I didnt follow this conversation (and I think i wont undestand the most) but can you implement a reference (variable?) in the core-revised cards to the "original" ones, so that we can "sort by set" with the card-positions of the pre-rotated cards? f.e. in card 20001 original_number=04041
Perhaps getting the card-data from there too(except scan)?

NoahTheDuke:
In this potential change, that issue would be resolved by having only a single copy of each card in the database, and then creating lists of those single cards in packs. Your issues will be allayed very soon!

eljenso:
Any news regarding this issue/PR?  
The change proposed in @NoahTheDuke's latest comment sounds promising.

Alsciende:
This PR is a proposed change to the data. Currently the application is not able to use that data. In the event where NRDB would have the same data model as 5RDB (where cards are separate from packs), this PR would be a good starting point for the new json repository.

Alsciende:
We'll come back to this PR and its discussion in the event the data model changes.

PR #260, opened 2017-10-03
Updating cycles for Oct 2017 rotation

jwarwick:
Marked `Core`, `Spin`, and `Genesis` as rotated in `cycles.json`

Also saw that there was a `rotations.json` file, should I be using that instead of the `rotated` flag? If so, let's remove the `rotated` field.

### DIFF
--- a/cycles.json
+++ b/cycles.json
@@ -10,14 +10,14 @@
         "code": "core",
         "name": "Core Set",
         "position": 1,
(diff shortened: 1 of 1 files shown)


gereons:
As long as there's no other rotation info available via the NetrunnerDB API, I'd say we keep the `rotated` flag.

PR #300, opened 2018-01-27
add spoilers

SuperFlamingTominoHead:


SuperFlamingTominoHead:
Hmm, so the two first check pointed out the errors, but the last check doesn't, just says there's a formatting error? Could it be problems with tab/space?

SuperFlamingTominoHead:
I guess it was...

Alsciende:
@SuperFlamingTominoHead "NGO Front" is missing a faction_cost.

gereons:
Fixed in #301

PR #302, opened 2018-01-28
add Kampala Ascendent

SuperFlamingTominoHead:


gereons:
adding just the packs reference makes the validation fail.

PR #322, opened 2018-03-06
Rar

gereons:


Alsciende:
Shouldn't it be in rotations.json too?

gereons:
You're right, see #323.
Wouldn't it be better to invert the logic for rotations, so that it only lists the stuff that's rotated out, instead of an ever-growing list of stuff that remains?

NoahTheDuke:
If anything, rotations.json should hold every "complete" version of Standard and Cache Refresh legalities, like how mwl.json holds all previous MWLs. (By complete, I mean listing all packs of a cycle, even if they're not released yet.)

gereons:
The problem I see with the current design is that whenever a new cycle comes out, we have to remember to edit `rotations.json` and add the new cycle to the (soon multiple) array(s) there. That smells like an unnecessary maintenance burden to me.

Alsciende:
Yes, it would make sense to list all "rotated out" cycles instead of all "allowed" cycles.

PR #326, opened 2018-03-12
new format + validation for `rotations.json`

gereons:
As discussed in #322, this PR changes the format of `rotations.json` (and adds it to the validation process) so that we only have to edit the file once for each rotation. Merging this depends on merging https://github.com/Alsciende/netrunnerdb/pull/209 in the netrunnerdb repo.


gereons:
We're not going to need this anymore 😭

PR #336, opened 2018-05-04
Shipment from Mirrormorph new text in Revised Core

JoelCFC25:
Never even noticed until today, randomly. I didn't include this, but maybe the errata tag could get taken off the Revised Core version of Punitive Counterstrike, since the card now matches.


gereons:
Nice catch! I just checked my old core cards, and in fact this text change was made between reprints.

PR #342, opened 2018-06-12
Fixed influence cost of NEXT Diamond

finnbar:
Image and cardgamedb says that NEXT Diamond costs 4 influence, data previously had it at 5.


Alsciende:
deployed, thanks

PR #347, opened 2018-06-19
Add xterm colours

ZoeB:
Translated by
[https://gist.github.com/MicahElliott/719710#file-colortrans-py](https://gist.github.com/MicahElliott/719710#file-colortrans-py). Some of these look a little off to me, but they're at least a good start.  (I've made a command-line-based deckbuilder, so this does have a practical, if very niche, application.)

### DIFF
--- a/schema/faction_schema.json
+++ b/schema/faction_schema.json
@@ -10,6 +10,10 @@
             "minLength": 6,
             "type": "string"
         },
+        "color_xterm": {
+            "minimum": 1,
+            "type": "integer"
+        },
         "is_mini": {
             "type": "boolean"
         },
--- a/schema/faction_schema.json
+++ b/schema/faction_schema.json
@@ -6,11 +6,13 @@
             "type": "string"
         },
         "color": {
+            "description": "Faction color in hex triplet format.  This can be used to display faction-specific information in the appropriate color on an HTML webpage.",
             "maxLength": 6,
             "minLength": 6,
             "type": "string"
         },
         "color_xterm": {
+            "description": "Faction color in xterm format.  This can be used to display faction-specific information in (the closest match to) the appropriate color in an X11-compatible terminal window.",
             "minimum": 1,
             "type": "integer"
         },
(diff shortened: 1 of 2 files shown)


Alsciende:
Lol that's crazy, but in a good way. Can you add a "description" attribute to the "color_xterm" property to explain briefly what it means, how to use it?

schema/faction_schema.json
```diff
@@ -10,6 +10,10 @@
             "minLength": 6,
             "type": "string"
         },
+        "color_xterm": {
+            "minimum": 1,
```
Alsciende: please add a description for this property

ZoeB:
Sure, added.  How's that?  I couldn't see any existing description attributes anywhere, so had to take a bit of a guess as to the level of detail to go into.  (Really, the existing "color" property should probably be called something more specific too, like "color_hex", but I don't want to break backwards compatibility, so I left it alone.  I added a description for that colour type too, though.  I hope that's OK.)

If you'd like to see an example of this kind of thing, then the command line deckbuilder I'm working on prints out errors in red: https://github.com/ZoeB/noise/blob/0fcefb15b4babd4b6e3aef2ee14aba1b51526d3f/noise.py#L36  Using the same kind of escape sequence, and the colours to hopefully be specified in the .json file, it will similarly be possible to print out anything faction-specific in the appropriate colour.

Cheers!

Alsciende:
@ZoeB Alright, so using the escape sequence `\0202[0;31m` would print in Anarch color?

ZoeB:
Almost.  It took me a little bit of dissecting colortrans.py to figure out the exact code, but [here's an example](http://bytenoise.co.uk/misc/netrunner-faction-colours.png).

Try this in Python 3, in a compatible terminal (I'm using OS X myself):

```
print('\033[38;5;202mAnarch')
print('\033[38;5;62mCriminal')
print('\033[38;5;77mShaper')
print('\033[0m')
```

ZoeB:
[Thank you](http://bytenoise.co.uk/misc/netrunner-faction-colours-2.png)! ❤️

PR #353, opened 2018-07-08
Use code of cycle instead of pack for RaR

eljenso:
Other deluxe cycles are also declared via cycle code here.


gereons:
@Alsciende is this maybe the reason why so many new published decks are still listed as "pre-rotation"?

PR #365, opened 2018-09-11
add cards from Magnum Opus

gereons:


NoahTheDuke:
Thanks for doing this, Gereon! I like how you included the community and multiplayer cards.

NoahTheDuke:
Might want to change the listed cards, if you're going to include them in rotations, cuz the multiplayer ID isn't legal.

gereons:
Hm. You're right. We'll need another cycle and pack just for the NAPD Multiplayer id. I'll amend the PR later today.

PR #369, opened 2018-11-19
System Core 2019

NoahTheDuke:
Sorry about the wait, y'all. MWL up tomorrow.


gereons:
Thanks, Noah! I'll take a look later today.

gereons:
I believe we also need to update `rotations.json`. I'll post a separate PR for that.

PR #375, opened 2018-11-27
Additional Card Validation

NoahTheDuke:
After #372, I realized that more could be done to strengthen the card validation schema. So after reading the json-schema documentation, I've gotten this far.


binarymelon:
Actually the schema link may not be correct.  Currently away from a computer with internet, but the validation script uses a draft 4 validator.  I'm assuming the link provided is the most recently released draft (7)?

NoahTheDuke:
Good point, @binarymelon. The Python library used currently can't even verify against Draft 6 or 7, so it's best to keep these parts in sync.

PR #377, opened 2018-12-07
Added Astro errata

x3N1GM4x:
Added errata from FAQ 3.1 to Astroscript Pilot Program (Limit 1 per deck). With Eternal format out there now, it's probably worth fixing this.

In the process, I seem to have broken my local dev environment, so cannot fully test that this works as expected - I can see that the errata show on the card page as expected, but I didn't get a chance to test whether the cards is now limited to a quantity of 1 on the Build tab of deckbuilder, so if someone could give this a pull and check before merging it, that would be great!

Pending successful tests, I believe this resolves [this issue](https://github.com/Alsciende/netrunnerdb/issues/29) from the netrunnerdb repo.

### DIFF
--- a/pack/core.json
+++ b/pack/core.json
@@ -1463,10 +1463,10 @@
         "position": 81,
         "quantity": 2,
         "side_code": "corp",
-        "text": "Place 1 agenda counter on AstroScript Pilot Program when you score it.\n<strong>Hosted agenda counter:</strong> Place 1 advancement token on a card that can be advanced.",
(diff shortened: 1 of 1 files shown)


gereons:
Setting `uniqueness` to false is neither correct nor necessary, afaict. The deck builder does not pay attention to it, since uniqueness is a play-time effect, and might even display the card's name incorrectly with a diamond where none should be. Instead, it should look at `deck_limit` to determine how many copies area allowed in a deck,

PR #378, opened 2018-12-08
set APP's deck_limit

gereons:
see #377


gereons:
It just occurred to me that that we changed APP back in February, see #312. The question is, do we (or NISEI) want to keep the old erratum for APP around, even when it was removed from the official FAQ/Errata documents?

PR #384, opened 2019-02-09
Add Flavor and Illustrator from previous version to sc19 cards

NoahTheDuke:


NoahTheDuke:
Nah, just a quick look to make sure I didn’t eff anything up!

PR #385, opened 2019-02-19
Ashes: Cycle and Packs, with The Nihilist, The Class Act, The Artist

NoahTheDuke:
~Still need the cycle.json and pack.json files, but this is a good start.~ Done!

### DIFF
--- a/packs.json
+++ b/packs.json
@@ -9,6 +9,15 @@
         "size": 20
     },
     {
+        "code": "ad",
+        "cycle_code": "ashes",
+        "date_release": "2019-03-15",
+        "ffg_id": null,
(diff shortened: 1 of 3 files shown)


packs.json
```diff
@@ -8,6 +8,15 @@
         "position": 1,
         "size": 20
     },
+    {
+        "code": "ad",
+        "cycle_code": "ashes",
+        "date_release": "2019-03-15",
```
x3N1GM4x: According to [this](http://nisei.net/article/HYPER01), Downfall is released 18th March, not 15th.
```suggestion
        "date_release": "2019-03-18",
```

PR #387, opened 2019-02-20
Nanoetching Matrix and Lucky Charm

NoahTheDuke:
Updates from http://nisei.net/article/2019AScoop2


NoahTheDuke:
NetrunnerDB uses strong in other flavor texts, so I followed suit.

x3N1GM4x:
The strange thing now that I actually look at the spoiled card image is that all of the flavour text on The Nihilist is italicised *except* for the `I'm` - maybe one to just get some clarification from Creative on, but almost certainly not a big deal in any case. 👍

PR #393, opened 2019-02-23
update card positions according

gereons:
initially posted numbers have changed, according to https://www.reddit.com/r/Netrunner/comments/atkwd3/downfall_number_crunch/eh39qoz/

I didn't reorder the cards within the file for now, and more importantly did not change the codes, as they're already published and in use, we'll have to keep this in mind when assigning new codes for the rest of the cards.


NoahTheDuke:
Last night, I actually manually changed it all and then went through and updated all of the decks that were using the previewed cards. I forgot to put my changes into a PR, cuz it was like 5 am (I was up for other reasons lol), but the database and card data on nrdb matches the new numbers, with no duplicates

gereons:
Updated the PR. I would ask that we refrain from these kinds of hacks in the future, as Netrunnerdb isn't the only place where card codes are used as primary keys. Changing these after publishing them needlessly invalidates decks built elsewhere, like the Android Deckbuilder and Net Deck.

NoahTheDuke:
I’m sorry for not speaking with you first. That was hasty and rude of me. I’ll be sure to discuss changes like this in the future before acting.

gereons:
It's all good, no apologies necessary :)

No one is really building serious decks anyway with the new cards at this stage, but the more cards are published the more important this becomes - maybe we should stop assigning pseudo-meaningful codes entirely.

gereons:
Noah, can you merge this? I've got the next set of scoops already lined up...

NoahTheDuke:
Sorry! Been away from the computer.

PR #394, opened 2019-02-26
Rezeki

gereons:


x3N1GM4x:
This has made it into production, but we shouldn't be publishing the Community Spoilers until the week is over - can we retract for now and publish again on Monday perhaps?

Also, there's no image for Rezeki in production.

NoahTheDuke:
We can’t remove the cards once they’re live, but leaving the card images off is fine.

gereons:
Why should we not publish the data anyway? It's public knowledge already, available on Facebook, reddit, Youtube, imgur and who knows where else, but not where people can actually use it? That seems weird to me. It's also contrary to how spoilers have been treated in the past, namely that they were made available ASAP.

x3N1GM4x:
Hi @gereons - I totally understand where you're coming from and for spoilers issued directly by NISEI, adding them immediately to NRDB is totally fine (and in fact preferable!).

The difference here is that Content Creators involved in the community spoilers this week were promised a short period of exclusivity on the cards from NISEI, to help drive people interested in the spoilers primarily to their channels/sites, to make the spoiler publicity beneficial for those content creators as well as for NISEI.

This kind of short-term exclusivity on spoilers may or may not happen again in the future, but we should certainly be wary of deploying spoiled cards to production. There was absolutely no problem with you submitting this PR (and the one for Rezeki), we just shouldn't have deployed to production until next week.

I hope that makes sense, but please reach out if you want to discuss further - your contributions to NRDB are super appreciated! ❤️

gereons:
Even if there were such a thing as exclusivity for this kind of data in today's internet (which I sincerely doubt, and the copy-pasting that's been going on with the spoiled cards shows that it didn't work), I don't think that withholding the data from the one place where people can do more than just look at it is useful or beneficial to anyone.

Case in point: The [Trace 5 video](https://www.youtube.com/watch?v=tOiH7QVqmUQ) has had >1300 views within 18 hours, which is the highest view frequency we've had in a long time. I have a hard time imagining that the cards being spoiled elsewhere has any measurable impact on the number of views.

x3N1GM4x:
I appreciate that and I'm not saying I disagree with you on the logic of the exclusivity, but the point is that it was promised to those content creators and in the interest of trust between NISEI and those content creators it should be honoured this time.

I have been made aware of instances where content creators have been disappointed to see the cards they scooped being published elsewhere without any link back to their original scoop content, so we know this does matter to people.

We absolutely can't stop other people from spreading the scoops, but that doesn't mean we should go back on what has been promised to content creators in this instance.

Trust me though, when discussions begin around future community scoops of this nature, I fully intend to make sure everyone involved understands the argument against promising this kind of exclusivity again.

PR #395, opened 2019-02-26
trace 5 scoops

gereons:
from https://www.youtube.com/watch?v=tOiH7QVqmUQ


x3N1GM4x:
Seems this has somehow made it into production as these cards are now visible on NRDB, when they should not be until Community Spoilers week is over.

I don't want to potentially make things worse by trying to poke around in production and fix this myself, but @NoahTheDuke, could you take a look when you're next online? Thanks.

PR #397, opened 2019-02-27
scoops from CodeMarvelous' stream

gereons:
Turns out Rezeki isn't at position 26 after all, so I tentatively moved it to 25, while keeping the codes intact.


NoahTheDuke:
> Turns out Rezeki isn't at position 26 after all, so I tentatively moved it to 25, while keeping the codes intact.

Ugh, that's so annoying. I think you've done it right. Almost makes me wish we didn't allow deck building and reviews for unreleased cards. ;)

gereons:
It's only annoying because we have these horrible pseudo-meaningful identifiers. If Rezeki's code were a UUID or an auto-incemented serial no. (as it should be), all would be good...

Also, there's precedent in the RCS cards where the codes don't match the pack/position schema.

PR #410, opened 2019-03-08
Fixed typo in artist name

x3N1GM4x:


x3N1GM4x:
I've also replaced the spoiler image with the updated one in production, so this just needs to be copied to the main images folder at the same time as deploying this change.

x3N1GM4x:
Deployed to production, with updated image.

PR #411, opened 2019-03-08
Fix Baklan Bochkin text error

NoahTheDuke:
Small error!


NoahTheDuke:
i'll give this one to @gereons with #412!

PR #419, opened 2019-03-11
Added remaining runner spoilers

x3N1GM4x:
http://nisei.net/article/2019AScoop10


pack/df.json
```diff
@@ -1,4 +1,21 @@
 [
+    {
+        "code": "26001",
+        "cost": 2,
+        "deck_limit": 3,
+        "faction_code": "anarch",
+        "faction_cost": 1,
+        "flavor": "With each passing day alone, hoshiko found it harder to think. With each hour, the static grew louder.",
```
NoahTheDuke:
```suggestion
        "flavor": "With each passing day alone, Hoshiko found it harder to think. With each hour, the static grew louder.",
```

PR #420, opened 2019-03-12
Fixed minor card text issues to match newly-provided card images

x3N1GM4x:
As per updated card images provided by NISEI Creative team (which I've already deployed to production).


x3N1GM4x:
Deployed to production.

PR #425, opened 2019-03-15
Fixed typo on Utae

x3N1GM4x:
Once per run instead of once per turn.


NoahTheDuke:
Does it need a period at the end of sentence?

x3N1GM4x:
Deployed to production.

PR #426, opened 2019-03-17
Non-icebreakers can't have null strength

NoahTheDuke:
Removes `null` strength from Trypano and eXer.


x3N1GM4x:
Deployed to production.

PR #428, opened 2019-03-19
Copy-editing text

SpencerDub:
The flavor text on Vulnerability Audit is inconsistent with the card image, and I'm not sure which is correct.

Otherwise, I believe flavor texts and rules texts are now all in line with card images.


pack/df.json
```diff
@@ -59,7 +59,7 @@
         "deck_limit": 3,
         "faction_code": "anarch",
         "faction_cost": 3,
-        "flavor": "\"Net space is an abstraction, a white lie protecting fragile comprehensions. Do not limit yourself.\" -z\\h/r",
+        "flavor": "\"Net <strong>space</strong> is an abstraction, a white lie protecting fragile comprehensions. Do not limit yourself.\" -z\\h/r",
```
x3N1GM4x: Won't using `<strong>` here bold the text, rather than un-italicise it like on the card image?

For example, check QPM:

https://github.com/Alsciende/netrunner-cards-json/blob/master/pack/dad.json#L114

pack/df.json
```diff
@@ -511,7 +511,7 @@
         "deck_limit": 3,
         "faction_code": "neutral-runner",
         "faction_cost": 0,
-        "flavor": "I didn't say your hopper should be hauling garbage. I said it should be hauled away as garbage.",
+        "flavor": "I didn't say your hopper should be hauling garbage. I said it should be hauled away <strong>as</strong> garbage.",
```
x3N1GM4x: Won't using `<strong>` here bold the text, rather than un-italicise it like on the card image?

For example, check QPM:

https://github.com/Alsciende/netrunner-cards-json/blob/master/pack/dad.json#L114

pack/df.json
```diff
@@ -672,7 +672,7 @@
         "deck_limit": 3,
         "faction_code": "haas-bioroid",
         "faction_cost": 3,
-        "flavor": "The Net abhors a vacuum. Any unexplained gap must be made.",
+        "flavor": "The Net abhors a vacuum. Any unexplained gap must be <strong>made.</strong>",
```
x3N1GM4x: Won't using `<strong>` here bold the text, rather than un-italicise it like on the card image?

For example, check QPM:

https://github.com/Alsciende/netrunner-cards-json/blob/master/pack/dad.json#L114

pack/df.json
SpencerDub: The few references I could quickly find for unitalicized words in flavor text suggested they were rendered using `<strong>` in the card data. This obviously isn't a perfect translation, but in the absence of counterexamples, that's what I went with.

That said, if there is precedent for just unitalicizing those emphasized words in the card data, by all means, that makes more sense.

NoahTheDuke: What's a good alternative to `<strong>` in flavor? `<em>`?

SpencerDub: If we use `<em>`, will it properly display as unitalicized in the italicized flavor text? Is the NetrunnerDB CSS set up to handle that?

Because if it is, then totally, `<em>` makes sense. It also makes semantic sense.

pack/df.json
```diff
@@ -909,7 +909,7 @@
         "deck_limit": 3,
         "faction_code": "nbn",
         "faction_cost": 1,
-        "flavor": "Though many countries attempted to regulate digital loot boxes in the early 21st century, GameNET has managed to circumvent any such laws via explicit, transparent percentage rates...and some very determined lobbyists.",
+        "flavor": "Though many countries attempted to regulate digital loot boxes in the early 21st century, GameNET has managed to circumvent any such laws via explicit, transparent percentage rates... and some very determined lobbyists.",
```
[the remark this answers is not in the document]
SpencerDub: That's odd, because if there were no space after the ellipsis, I expect the line wouldn't break where it does on the card.

Also, I totally disagree with that from a stylistic perspective (ellipses are always followed by spaces!), but I suppose that's neither here nor there.

pack/df.json
x3N1GM4x: Yeah, I thought I'd already done at least one FT like this, but my understanding is that `<em>` would have the desired effect - I can test on my local later if someone else doesn't beat me to it.

pack/df.json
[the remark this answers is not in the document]
x3N1GM4x: Yeah, I had the same view, which is why I confirmed it with them, and they specified no space, so I guess we should take that as canon, even if we disagree. 😄

PR #431, opened 2019-03-25
Correct Hadrian's Wall flavor text

SpencerDub:
Capitalize "his" in second sentence

### DIFF
--- a/pack/core.json
+++ b/pack/core.json
@@ -1829,7 +1829,7 @@
         "deck_limit": 3,
         "faction_code": "weyland-consortium",
         "faction_cost": 3,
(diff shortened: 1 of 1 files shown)


x3N1GM4x:
Good catch, thanks!

x3N1GM4x:
Deployed to production.

x3N1GM4x:
Hmm, strange - this one definitely deployed:

image

But I'm not seeing the change in the front end like I am for the Masvingo one which went in at the same time. Can anyone else confirm what they are seeing?

SpencerDub:
It's live on the data for the original core, but if you view the data for Revised Core or System Core 2019, it doesn't show. That was my mistake.

PR #433, opened 2019-03-25
Fix Hadrian's Wall flavor text capitalization

SpencerDub:
Fixed this in Original Core, but the issue was present in the Revised Core and System Core 2019 data too. This fixes System Core 2019.


x3N1GM4x:
Deployed to production.

PR #437, opened 2019-03-31
Remove additional line breaks, arrange magnum opus by code

NoahTheDuke:
Other packs are arranged by code, so should Magnum Opus. It changes nothing in the data itself, just the order of the cards in the file. (Why is the position different than the numerical order?)

Najja and IPO shouldn't have two linebreaks, as that's not how it's been done _anywhere_ else.


x3N1GM4x:
Deployed to Production.

PR #438, opened 2019-04-05
Order Revised Core and Magnum Opus by position, not code

NoahTheDuke:
I don't know why the "position" of the Magnum Opus cards doesn't match the printed numbers on their cards. That should be fixed, but that can happen in a different PR.

This is not a functional change, just a sorting of the packs by `position`. I used a script to automate this; all other packs are currently sorted properly.


gereons:
I think positions needed to be in order before https://github.com/Alsciende/netrunnerdb/pull/331 landed. We can make them match the printed number now, afaict.

NoahTheDuke:
Cool, good to know. I'll see if things still work locally before opening a PR.

PR #440, opened 2019-04-24
Fix various typos and missing FT

SpencerDub:

PR #442, opened 2019-04-26
Fix various typos and missing FT

NoahTheDuke:
Re-up of #440, rebased on master. Sorry about that, @SpencerDub! But it's all worked out!


NoahTheDuke:
Pushed to prod

PR #443, opened 2019-04-26
Fix small typos in RAR

SpencerDub:
* Place all of Mache's cost within `<strong>` tags
* Correct missing word in Too Big to Fail's flavor text
* Add missing period to end of Otoroshi's flavor text


NoahTheDuke:
You're crushing it, dude. Thanks so much!

PR #445, opened 2019-05-01
MWL 3.2

gereons:


gereons:
Indeed, I missed GPI. Good catch!

Re Aesop: no idea. I started by copying the list from 3.1 where it's exactly the same, so I assume this just works as-is?

x3N1GM4x:
Aha, fair enough, if it ain't broke, don't fix it! :p

PR #447, opened 2019-05-11
Add missing versions

NoahTheDuke:
Each of the included entries is a different printing/version than the one at release, which is why it's not included. The cards are Eli 1.0, Aesop's Pawnshop, and Magnum Opus.


NoahTheDuke:
There should be no functional changes here, so I expect we're in the clear!

PR #454, opened 2019-07-19
Added Interns errata from 2nd FFG printing

x3N1GM4x:
I already SCPed an updated card image onto production, so this just makes the text match the image now.


x3N1GM4x:
Deployed to production.

PR #455, opened 2019-07-25
Added Crisium errata from CR 1.3

x3N1GM4x:


pack/fc.json
```diff
@@ -142,7 +142,7 @@
         "position": 48,
         "quantity": 3,
         "side_code": "corp",
-        "text": "Successful runs against this server are not considered to be successful or unsuccessful for the purposes of card abilities.\nLimit 1 <strong>region</strong> per server.",
+        "text": "“Runs against this server cannot be declared successful. (This effect does not cause runs to become unsuccessful.)\nLimit 1 <strong>region</strong> per server.<errata>Errata from CR 1.3</errata>",
```
NoahTheDuke:
```suggestion
        "text": "Runs against this server cannot be declared successful. (This effect does not cause runs to become unsuccessful.)\nLimit 1 <strong>region</strong> per server.\n<errata>Errata from CR 1.3</errata>",
```

x3N1GM4x: ooops, good catch!

x3N1GM4x:
Looks like Travis got stuck while booting its VM and the CI build is just not gonna start, but I can't see how to force it to try again... Anyone?

NoahTheDuke:
Deployed to production

x3N1GM4x:
Not seeing this on the production site even after a hard refresh, am I being impatient? 😝

x3N1GM4x:
Oh, I am so dumb. I only updated the original card, not the one in SC19. 🤦‍♂ 
I'll drop another PR.

PR #460, opened 2019-08-22
Added Swift (newly spoiled Crim Console) and Uprising pack

x3N1GM4x:
I've already uploaded the spoiled full bleed card image to production, so it's ready to go...

I put a placeholder release date for Uprising of end of the year, although we expect it to be before NYE itself!


NoahTheDuke:
Deployed

PR #464, opened 2019-09-25
Add Magnum Opus Reprint

NoahTheDuke:
Also fixes minor wording for Slot Machine and Timely Public Release, and updates the MWL to handle the new cards.


x3N1GM4x:
Deployed to production.

PR #473, opened 2019-11-28
Added both new interrupt cards from 28/11 spoiler article

x3N1GM4x:


x3N1GM4x:
I'll also need to add the interrupt glyph, will either add that to this PR shortly, or log a separate one.

x3N1GM4x:
I won't deploy this until I sort out adding the interrupt glyph, obv.

x3N1GM4x:
Deployed to production.

PR #492, opened 2019-12-11
Minor corrections to Uprising cards

SpencerDub:
Replaced a `<strong>` tag in Buffer Drive's flavor text with `<em>`.

Added `<strong>` tags around card subtypes in text of Afterimage, Euler, Penrose, DreamNet, etc.

Fixed recurring credit in Mantle.

Corrected an error in The Back's text ("each run" ➡️ "each turn") and bolded the non-click text in the paid ability's cost.

Other general copy-editing (periods, word order).

### DIFF
--- a/pack/ur.json
+++ b/pack/ur.json
@@ -210,7 +210,7 @@
         "quantity": 3,
         "side_code": "runner",
         "strength": 2,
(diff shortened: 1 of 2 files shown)


NoahTheDuke:
Love your tireless work.

PR #497, opened 2019-12-14
Fix em dash in Aniccam FT

SpencerDub:


NoahTheDuke:
I don't believe in em dashes but I guess I'll allow it.

PR #500, opened 2019-12-16
Fixed artist name mistakes in Uprising

x3N1GM4x:


x3N1GM4x:
Deployed to production.

PR #505, opened 2019-12-17
Rotated C&C and Lunar

x3N1GM4x:
Who knew? This is done here and not in NRDB itself based on the `rotations.json`  🤷‍♂

### DIFF
--- a/cycles.json
+++ b/cycles.json
@@ -24,7 +24,7 @@
         "code": "creation-and-control",
         "name": "Creation and Control",
         "position": 3,
-        "rotated": false,
+        "rotated": true,
         "size": 1
     },
     {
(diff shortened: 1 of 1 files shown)


x3N1GM4x:
Deployed to production.

gereons:
Guys, it's not the 27th yet...

x3N1GM4x:
> Guys, it's not the 27th yet...

Dang, you caught us! As NISEI have updated nisei.net to show rotation already and the general feedback I've seen is that people are treating rotation as if it has happened even prior to it becoming official on the 27th, it seemed to be fine to do this now.

Besides not being technically correct, does this cause any problems you can see @gereons? We could easily have missed something important! 😄

gereons:
Nah, this is strictly about being technically correct :)

x3N1GM4x:
Then I'll defer discussing it for...I dunno...9 more days. 😝

PR #508, opened 2019-12-23
Fix Ikawah FT

SpencerDub:
Deleted an extra word


x3N1GM4x:
Deployed to production, thanks as always for the catches @SpencerDub!

PR #513, opened 2020-01-02
Slot machine edit

x3N1GM4x:
Name update for Jess Horig on Slot Machine.


x3N1GM4x:
Deployed to production.

PR #514, opened 2020-01-05
Update the card schema to have explicit storage for flip cards.

plural:
Populate this data structure for all of the flip identities.

This doesn't affect NRDB in any way yet.  Feedback desired on the format!


pack/ur.json
```diff
@@ -5,6 +5,17 @@
         "deck_limit": 1,
         "faction_code": "anarch",
         "flavor": "Please, let me have this dream.\nFlip side:\nI'm going to be my own kind of hero.",
```
benhamill, line 7: Probably the flavor should be edited to remove the "flip side" notation?

plural: it should, but I am going to hold off on that until NRDB can handle the flip side data.

If i make this change right now, this text will simply disappear from NRDB, sadly.

plural:
I'm closing this out (finally).

We'll handle flip cards properly in the new format for the new API/app, but leave things alone for the original files.

PR #515, opened 2020-01-10
Updated card text for Daily Casts in Uprising

x3N1GM4x:
As spotted by Jamie from Rules, the card text for the Uprising version of Daily Casts was not updated.


x3N1GM4x:
Deployed to production.

PR #516, opened 2020-01-11
Clean up MWL 3.4b, Daily Casts text in Creation & Control

NoahTheDuke:


pack/cac.json
```diff
@@ -957,7 +957,7 @@
         "position": 53,
         "quantity": 3,
         "side_code": "runner",
-        "text": "Place 8[credit] from the bank on Daily Casts when it is installed. When there are no credits left on Daily Casts, trash it.\nTake 2[credit] from Daily Casts when your turn begins.",
+        "text": "When you install this resource, load 8[credit] onto it. When it is empty, trash it.\nWhen your turn begins, take 2[credit] from this resource.",
```
x3N1GM4x: So, do we usually update the card text on an older printing of a card rather than leaving it matching the printed card? I wonder if it makes more sense to have ti so if people are viewing the C&C version of the verd, they see the C&C printed text, with the updated Uprising text only showing when they're viewing that printing of the card?

gereons: Good point. For other reprints in uprising we have in fact changed all copies of those cards to match the new text, e.g. with False Lead. I think its best to do this for all cards.

x3N1GM4x: Hmmm, I guess it depends what we think the card text is supposed to represent. For me, it seems sensible to replicate what exists in the real world, where there's a C&C version of the card with it's own text, which has now rotated, then an Uprising version with the updated text, which is currently legal. I Mean, we show the appropriate images against each printing of a card, why not the text as well?

gereons: IMHO, the card text should accurately represent what the card does in-game. Also, there's Eternal where old and new cards are perfectly legal at the same time. 

Ideally, the text changes including errata would be photoshopped back into the images, but AFAIK noone has bothererd to do that except for Astroscript.

NoahTheDuke: Yeah, every other card, those with errata or with changed printed text, have had all of their versions changed.

x3N1GM4x: Okey doke, I personally disagree with that being the right way to go, but sounds like I'm outnumbered, so happy to go with the consensus.

x3N1GM4x:
Deployed to production.

PR #517, opened 2020-01-14
Fixed issue where card ability had been entirely overwritten

x3N1GM4x:
@NoahTheDuke took the errata too literally:

```
11.2.6. Zaibatsu Loyalty (Original Core 71)
a. Should read: “[interrupt]– When a card would be exposed, you may rez this asset.”
```
😛 😝


x3N1GM4x:
Deployed to production.

PR #528, opened 2020-07-07
s/-/–/ Replace hyphen with en-dash on interrupt text.

plural:
at Jamie's request.


gereons:
Unsure why the validator script complains - this looks like a perfectly valid UTF-8 character to me, and other JSON parsers I've used agree...

NoahTheDuke:
Let me take a look at the validator.

NoahTheDuke:
it's line 58 or something in `validate.py`. `.replace(u"\u2013", "-")` is what yells when you put an en-dash in.

gereons:
Wow, that's bad. I wonder why all this massaging of the `json.dumps` output seemed necessary... Does the NRDB importer still read the files OK when you get rid of that cruft? If so, I'd vote for nuking it.

plural:
oh, looks like i didn't read very carefully.  Do we want to yank those parts of the validator as a separate pull request?  I don't understand why we would single out those specific characters either.

NoahTheDuke:
> Wow, that's bad. I wonder why all this massaging of the `json.dumps` output seemed necessary... Does the NRDB importer still read the files OK when you get rid of that cruft? If so, I'd vote for nuking it.

That stuff makes sure people are formatting the json the same way. I don't want to have PRs with tabs and hanging spaces and stuff, which this handles. I suspect removing just that line would be fine, but I also things changing it so interrupts are hardcoded to be allowed would be smarter in the short-term. I don't know what the best long-term solution is.

Maybe something like the below would be best for now:
```    formatted_data = (
        formatted_data
        .replace(u"\u2018", "'")
        .replace(u"\u2019", "'")
        .replace(u"\u2212", "-")
        .replace(u"\u2013", "-")
        .replace(u"[interrupt] -", "[interrupt] \u2013")
        .replace("\\r\\n", "\\n")
        .replace(" \\n", "\\n")
    )```

gereons:
> That stuff makes sure people are formatting the json the same way  

Sure, but wouldn't just using `json.dumps` be sufficient to ensure that, plus maybe the line endings / trailing blanks stuff? Specifically disallowing certain kinds of quotation marks and dashes seems really weird to me.

NoahTheDuke:
> Sure, but wouldn't just using `json.dumps` be sufficient to ensure that, plus maybe the line endings / trailing blanks stuff? Specifically disallowing certain kinds of quotation marks and dashes seems really weird to me.

Yeah, I don't know exactly why this was done. Unifying/formalizing these aren't outrageous to me but I also don't think we need them anymore.

PR #532, opened 2020-09-27
make Mâché unique

gereons:
see https://github.com/Alsciende/netrunnerdb/issues/451


gereons:
Bildschirmfoto 2020-09-27 um 11 23 29

So the check's result still seem to be getting lost somewhere - I would suggest that until this is fixed we turn the requirement of a passing Travis test off. See https://github.com/Alsciende/netrunner-cards-json/pull/530#issuecomment-692188013 for an alternative.

plural (approved):

PR #533, opened 2020-09-28
Add Ban list 20.09

plural:
Add Standard Ban list 20.09
+SSL Endorsement
+Gold Farmer
+PAD Tap


plural:
closing out in favor of the original that gereons made for us!

PR #538, opened 2020-12-12
add Salvaged Memories

gereons:
Adds the "Salvaged Memories" digital-only pack from https://nisei.net/article/salvmem


NoahTheDuke:
Hey thanks for tackling this! To double check (cuz I don't have my computer this weekend and em checking on my phone is hard), the cards are in faction and type and name order, right? So anarch, criminal, shaper, hb, jinteki, nbn, weyland. And then for runner: identity, event, hardware, program, resource. And corp: identity, agenda, asset, ice, operation, upgrade. And then alphabetical by name after that.

If that's how you have it, great, I'll approve!

gereons:
Hm, lemme check. I copy-pasted them in the order they were listed in the NISEI article, so at least they're sorted by faction.

gereons:
Two or three were out of order (by type), but this should fit now.

NoahTheDuke (approved):

PR #555, opened 2021-03-12
Allow the code to run from other package.

plural:


src/index.ts
```diff
@@ -6,7 +6,9 @@ function makeReader(filename: string) {
 
 	return function(): Record<string, any>[] {
 		if (json.length === 0) {
-			json = JSON.parse(fs.readFileSync(resolve('.', filename + '.json'), 'utf-8'));
+			let path = resolve(__dirname, '..', filename + '.json');
+			console.log('path is ' + path);
```
NoahTheDuke, line 10: We should probably use a logging tool here, so consumers of the API don't have messages they don't want.

src/index.ts
```diff
@@ -19,11 +21,15 @@ export const getMwlJson = makeReader('mwl');
...
+export function getDirName(): string {
+		return resolve(__dirname, '..', 'pack');
+}
+
 export function getCardsJson(): Record<string, any>[] {
 	const CARDS_JSON: Record<string, any>[] = [];
 
 	if (CARDS_JSON.length === 0) {
-		const directory = resolve('.', 'pack');
+		const directory = resolve(__dirname, '..', 'pack');
```
NoahTheDuke, line 32: Use `getDirName` here?

src/index.ts
NoahTheDuke: the package `loglevel` is really good for this, tbh

plural: yeah, this was just me being sloppy at the end of the day.  This and the other method were there just to confirm what i thought was happening in the other repo.

src/index.ts
plural: deleted the function.

NoahTheDuke (approved):

PR #558, opened 2021-03-13
A couple of cards were missing stripped_text.

plural:
This caused errors in the importer for NRDB.


NoahTheDuke:
Stripped text should be nullable. If it's not, that's something we gotta fix in nrdb. These two cards don't have text fields so they shouldn't have stripped text.

plural:
yep, fixed in the importer instead of here, so i'll sacrifice this one to Apex.

PR #564, opened 2021-03-23
fix deck limit for SG 3/2 agendas

francescopellegrini:
I've just noticed this bug and wanted to contribute =)


NoahTheDuke:
Thank you! I totally forgot about this field haha.

NoahTheDuke (approved):

PR #568, opened 2021-03-26
Fix campaign cards

### LINKED ISSUE #567: Security Leak and Net Watchlist under wrong side
Side_code is switched for these 2 tdc currents

gereons:
fix Security Leak and Net Watchlist from tdc, make validator.py catch these mixups

fixes #567


NoahTheDuke (approved):

plural:
yay!  Thanks for the validator update, @gereons !!!

PR #574, opened 2021-04-02
Mark Honor & Profit and SanSan cycles as rotated

NoahTheDuke:


plural:
why do we have 2 rotated indications in the JSON?  Rotation doesn't take affect until the 9th, so i'd prefer to not submit this until then.

Ideally, we should clarify the duplicated source of truth here.

gereons:
> Ideally, we should clarify the duplicated source of truth here.

Absolutely. On that note, I made a suggestion to invert the logic related to `rotations.json` ages ago - right now, this file lists the valid packs after each rotation instead of what packs actually rotate in each rotation, that seems weird to me, and also entails editing every rotation record when a new pack comes out. 

The changes to the file and the database schema were pretty simple, I still might have those somewhere if you're interested.

NoahTheDuke:
That would be great, thank you!

gereons:
OK, found my old PRs from March '18: https://github.com/NetrunnerDB/netrunnerdb/pull/209 and https://github.com/NetrunnerDB/netrunner-cards-json/pull/326. I'll go see if I can update them to the current state of affairs... :-)

gereons:
It just occurred to me that SC19 and Salvaged Memories also need to be marked as rotated.

plural (approved):

PR #575, opened 2021-04-03
New rotations format

gereons:
This inverts the logic of `rotations.json` - instead of listing what's legal after this rotation, we now specify what's rotating out.

This PR requires a db schema change in NetrunnerDB, see https://github.com/NetrunnerDB/netrunnerdb/pull/512

### DIFF
--- a/rotations.json
+++ b/rotations.json
@@ -1,85 +1,40 @@
 [
-  {
-    "code": "rotation-2017",
-    "cycles": [
-      "creation-and-control",
-      "honor-and-profit",
-      "lunar",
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
Should we remove `"rotated"` from cycles.json?

gereons:
Yes, that would be the next logical step, but I didn't want to touch too many things at once. This is mostly because this also requires some code changes in NRDB, as we need to make sure that the `rotated` property in the public NRDB API stays intact, without being backed directly by a db column.

plural (approved):

PR #577, opened 2021-04-10
Mark sc-2019 and salvaged-memories as rotated.

plural:
thanks to @gereons for pointing this out.  I updated the server manually to unban all the ids. hehehehehe  lolsob

### DIFF
--- a/cycles.json
+++ b/cycles.json
@@ -136,7 +136,7 @@
         "code": "sc19",
         "name": "System Core 2019",
         "position": 25,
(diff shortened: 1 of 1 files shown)


NoahTheDuke (approved):

NoahTheDuke:
Do you gotta remove them from formats or something too?

plural:
the formats are a bit of a hack.  This does the right thing on the collections tab already if you click Standard, so i think we are good there.

PR #584, opened 2021-08-08
Add a subtypes banned list to each banlist object.

plural:
step 1 of simplifying the "currents are banned" handling for NRDB.

This plus a couple of NRDB changes will mean i won't forget during the next banlist and accidentally list out every single current again.  :D

### DIFF
--- a/mwl.json
+++ b/mwl.json
@@ -1461,6 +1461,9 @@
         "name": "Standard MWL 3.4b"
     },
     {
+        "banned_subtypes": [
+            "current"
+        ],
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Interesting idea! I'm not opposed. What about us using deck_limit instead of banned? Something we should think about moving away from to have unified language?

NoahTheDuke (approved):

plural:
Yeah, good call!  I moved this to mirror the cards list.  PTAL.

PR #590, opened 2021-09-20
Add tests for matching {stripped_,}text fields.

plural:
For any cards with multiple printings, ensure that the text &
stripped_text fields are identical for every version of the card.

This passes at head (huzzah!) and is a good, small bit of defense until we have card / printing files.

Example output:
Validation error for card 'Sure Gamble': Found 2 different versions of text.
    'Gain 9[credit].'
    'Gain 9[credit] if you are cool.'
Validation error for card 'Sure Gamble': Found 2 different versions of stripped_text.
    'Gain 9 credits.'
    'Gain 9 credits if you are cool.'
Found 0 formatting and 2 validation errors

### DIFF
--- a/validate.py
+++ b/validate.py
@@ -11,6 +11,8 @@ PACK_DIR="pack"
 SCHEMA_DIR="schema"
 formatting_errors = 0
 validation_errors = 0
+text_by_card_title = dict()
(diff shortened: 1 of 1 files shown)


NoahTheDuke (approved):
this is great! good catch.

PR #591, opened 2021-09-22
Update card text for 1.5 fixes.

plural:

PR #592, opened 2021-09-22
Fix arrows and dashes

NoahTheDuke:
Needs #591 to be merged first. I'll rebase on top once that's done.


plural (approved):
looks good.  maybe we should add a check in the validation script for the wrong forms in text and stripped_text in a followup?

NoahTheDuke:
Sure, I can add that!

validate.py
```diff
@@ -192,6 +192,54 @@ def validate_card(args, card, card_schema, pack_code, factions_data, types_data,
...
+            "Stripped text contains non-ascii characters in card: (pack code: '{}' title: '{}' stripped_text '{}')\n".format(
+                pack_code,
+                card['title'],
+                stripped_text,
+            ),
+            0,
+        )
+        validation_errors += 1
+
+def verify_text_has_fancy_text(args, card, pack_code):
```
plural, line 214: teehee

pack/draft.json
```diff
@@ -169,7 +169,7 @@
         "position": 13,
         "quantity": 1,
         "side_code": "corp",
-        "stripped_text": "Draft format only. If you have more cards rezzed than any other faction, when the Runner's turn begins, place an advancement token on a piece of ice.",
+        "stripped_text": "Draft format only. If you have more  cards rezzed than any other faction, when the Runner's turn begins, place an advancement token on a piece of ice.",
```
NoahTheDuke, line 172: oops, one sec

validate.py
```diff
@@ -192,6 +192,54 @@ def validate_card(args, card, card_schema, pack_code, factions_data, types_data,
...
+            args,
+            "Incorrect interrupt text in card: (pack code: '{}' title: '{}' text '{}')\n".format(
+                pack_code,
+                card['title'],
+                text,
+            ),
+            0,
+        )
+        validation_errors += 1
+    if ('Interface' in text) and (('Interface ->' in text) or ('Interface →' not in text)):
```
plural, line 230: why accept both formats instead of just the one?

plural (commented):
just one question

validate.py
NoahTheDuke: This is checking the "text" field, so it should be "Interface →" (with the unicode arrow). I'm raising an error if it's a dash and greater than or anything that's not a unicode error.

plural (approved):

PR #605, opened 2022-01-18
Add a set_types.json file and update validate.py

plural:
Starting to fill in the gaps for what is missing for NRDB v2 in the repo.


NoahTheDuke (approved):
This list of types is good? Hard to change once we release to the world lol

plural:
Well, we can label it private and subject to change.

I can solicit some feedback on it from other folks, but this is ok modulo
naming arguments. :)

PR #606, opened 2022-02-19
Add pack/set type values and update schema & test.

plural:
Part of prep-work for the new NRDB API.


plural:
this data is in the new printings.json file, so I will close this out.  Yay!

PR #610, opened 2022-02-28
Update deps to fix nanoid vulnerability issue

plural:


plural:
I'm making sure this is clean before I move it around and I dislike nags from dependabot.  :D

NoahTheDuke (approved):

PR #611, opened 2022-02-28
Fix busted tests, even though they are silly.

plural:


NoahTheDuke:
I don't know exactly what the existing error is, but we shouldn't be mixing assertion styles (`expect` vs `should`) and we should be using `import` syntax instead of CommonJS's `require` function.

NoahTheDuke (changes requested):
`should` -> `expect`

`require` -> `import`

plural:
I'm actually going to nuke this. I slept on it and the tests are too silly to fix. These as is are basically duplicating our schema validation.

PR #613, opened 2022-03-06
Add Midnight Sun Booster Pack cards.

plural:
i'm adding the new format JSON in a separate CL.


cycles.json
```diff
@@ -173,5 +173,12 @@
         "position": 31,
         "rotated": false,
         "size": 1
+    },
+    {
+        "code": "borealis",
+        "name": "Borealis",
+        "position": 32,
+        "rotated": false,
+        "size": 1
```
NoahTheDuke, line 182: Size 3, as midnight Sun and the other one are coming out later.

NoahTheDuke (approved):
This looks good to me!

cycles.json
plural: ah, thanks!

PR #617, opened 2022-03-11
Add an explicit subtypes JSON file.

plural:


NoahTheDuke (approved):
Mind formatting with spaces after the colon? Otherwise this is good.

PR #620, opened 2022-03-12
Updated json schema to latest, demonstrate validation with ajv

NoahTheDuke:
This is intended as a jumping off point for building something better. I had hoped to use `ajv` to handle all validation stuff, but it doesn't work well with arrays, only shows the first error instead of multiple errors lol. I think a smarter rewrite of the python validation script is good: instead of globals and printing, maybe an array of errors and pretty-printing them at the end? I don't know, I haven't explored that part because I had full faith that I could bang this out in a couple hours hahahaha

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,20 +9,21 @@
 			"version": "1.0.2",
 			"license": "ISC",
 			"dependencies": {
+				"ajv": "^8.10.0",
 				"lodash": "^4.17.21"
 			},
 			"devDependencies": {
 				"@tsconfig/recommended": "^1.0.1",
-				"@types/chai": "^4.2.15",
-				"@types/lodash": "^4.14.168",
-				"@types/mocha": "^8.2.1",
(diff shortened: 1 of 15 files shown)


plural:
what command do you run at the CLI to kick this off?

NoahTheDuke:
What do you mean?

plural:
I had to move this to tests/schema/validate ts so I could get this to run
when I ran npm test.  How were you running it?

Also, since I have you, I have a request for you on discord when you get a
moment. :)

NoahTheDuke:
I will pull out my laptop and check discord! I haven't gotten on a non-work computer since the new baby was born 💀.

I ran `npm test`. It's defined in `package.json` to run all test files in the `test/` directory, so in surprised you had to change it to another directory. I wonder what broke/changed.

plural:
big thanks to Noah for kicking this off.  I'm finishing this up in #656 .

PR #621, opened 2022-03-13
Set up Prettier pre-commit hook, format all JSON files with it.

plural:
This is step 1 in moving away from the python validator.

Removed the formatting code from the python validator.

Add a pre-commit hook that will update all JSON files when committing.

Run prettier over all JSON files, except the template/ files because they are not valid standalone JSON files.


NoahTheDuke:
This is great. Husky is very cool.

.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
NoahTheDuke, line 1: Any reason for this?

NoahTheDuke (approved):

.husky/.gitignore
plural: it happened when husky was added - i didn't add it explicitly.

/me does his own research

Looks like husky auto-gens it when installing: https://github.com/typicode/husky/issues/929

PR #634, opened 2022-05-19
Embolden costs in MSBP

SpencerDub:
Fixing bolding in costs in MSBP


distributive (approved):
Hakarl 1.0 also has a unbolded cost (the lose ability) if you want to have a look at that.

PR #635, opened 2022-05-20
Added card pools

distributive:
Card pools here represent the list of cycles, packs, and cards each format had at a given time, without the additional context of when they became legal or any banlists that applied to them.


card_pools/misc.json
```diff
@@ -0,0 +1,12 @@
+[
```
plural, line 1: i'm not sure these two make sense as card pools as is.

Draft would really be draft + anything, likely.

NAPD multiplayer would be the same.  

maybe just best to leave them out for now until we address draft deck building head on?

plural (approved):

PR #636, opened 2022-05-20
Added formats

distributive:
Formats here represent a series of snapshots for each official format, where each snapshot has a card pool, a start date, and an optional mwl entry. Exactly one snapshot should have "active" set to true. Snapshots should be ordered from old (first) to new (last).


plural (changes requested):
we can't delete rotations.json - it's used by the existing NRDB importer.

plural (approved):

PR #641, opened 2022-05-21
Update prettier to latest version.

plural:


NoahTheDuke:
I don't think this food what you think it does. If you want the latest version, you should have the latest version locked to a specific major.minor.patch. Otherwise, you'll get inconsistencies between users/CI.

PR #643, opened 2022-05-21
Javascript schema validation seeds

plural:
this is based on noah's PR from a while back.   If we like this, i'll make all the schema updates and remove the schema validation checks from the python app and set this up in our CI.


NoahTheDuke (commented):
Seems like a much better starting point than I had!

plural:
> Seems like a much better starting point than I had!

ha!

plural:
@distributive re: "remove the rotated flag" - yes.  I've got a PR in flight to make a clean break between v2 and v1 and i'll do that as part of that work.

PR #644, opened 2022-05-21
Add lint script for typescript.

plural:
Adding lint for typescript, which will become more relevant as we ditch python in this repo.

Fix the errors found by lint in 2 of the typescript files.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,5 +1,8 @@
 {
   "parser": "@typescript-eslint/parser",
+  "parserOptions": {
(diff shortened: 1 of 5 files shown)


.eslintrc.json
```diff
@@ -1,5 +1,8 @@
 {
   "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "warnOnUnsupportedTypeScriptVersion": false
```
NoahTheDuke, line 4: Why is this needed?

PR #650, opened 2022-06-05
Remove duplicate subtypes with the id/name fields flipped.

plural:


plural:
I'm not sure how i missed this originally, but it's fixed now, i guess.  :D

distributive (approved):

PR #652, opened 2022-06-12
Removed all instances of <i> and replaced them with <em>

distributive:
Some cards from Ashes use the wrong tag for italics. This reverts them to `<em>` to match the existing standard.


plural:
this is good, thank you!  Would you remove the pinhole threading commits from this PR, though, please?

PR #656, opened 2022-06-13
Create V2 JSON, move testing from python to typescript.

plural:
I have a couple types moved now, but not all of them, so i'll keep this draft until then.


distributive:
So is the plan in the long run to have two folders here for each version indefinitely?

plural:
Well, step 1 is separate and fix the tests, removing the python. Step 2 is generate v1 from v2.  It will be a while until we can remove the old format since we don't have a great handle on who else uses it and the biggest user (nrdb) isn't worth switching over, imo.

NoahTheDuke:
The list reporter can be configured in a .mocharc file, might help out.

distributive (approved):

PR #662, opened 2022-06-19
V2 expand schema

plural:
Add some more schemas.

Add formatting function for our general text to code/id transformation.

Update identities to have corp_identity and runner_id for card_type_id.

### DIFF
--- a/schema/v2/card_schema.json
+++ b/schema/v2/card_schema.json
@@ -57,8 +57,8 @@
               "enum": [
                 "agenda",
                 "asset",
-                "ice",
                 "corp_identity",
+                "ice",
                 "operation",
                 "upgrade"
               ]
@@ -74,9 +74,9 @@
               "enum": [
                 "event",
                 "hardware",
-                "runner_identity",
                 "program",
-                "resource"
+                "resource",
+                "runner_identity"
               ]
             }
           }
--- a/schema/v2/card_schema.json
+++ b/schema/v2/card_schema.json
@@ -1,0 +1,328 @@
+{
+  "$id": "https://github.com/netrunnerdb/netrunner-cards-json/blob/master/schema/v2/card_schema.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "additionalProperties": false,
+  "allOf": [
+    {
+      "title": "Faction id must be valid for the side.",
+      "type": "object",
+      "oneOf": [
+        {
+          "properties": {
+            "side_id": {
+              "const": "corp"
+            },
+            "faction_id": {
+              "enum": [
+                "haas-bioroid",
+                "jinteki",
+                "nbn",
+                "neutral-corp",
+                "weyland-consortium"
+              ]
+            }
+          }
+        },
+        {
+          "properties": {
+            "side_id": {
+              "const": "runner"
+            },
+            "faction_id": {
+              "enum": [
+                "adam",
+                "anarch",
+                "apex",
+                "criminal",
+                "neutral-runner",
+                "shaper",
+                "sunny-lebeau"
+              ]
+            }
+          }
+        }
+      ]
+    },
+    {
+      "title": "Type codes must be valid for the side.",
+      "type": "object",
+      "oneOf": [
+        {
+          "properties": {
+            "side_id": {
+              "const": "corp"
+            },
+            "card_type_id": {
+              "enum": [
+                "agenda",
+                "asset",
+                "ice",
+                "corp_identity",
+                "operation",
+                "upgrade"
+              ]
+            }
+          }
+        },
+        {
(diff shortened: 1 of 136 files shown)


schema/v2/card_schema.json
```diff
@@ -0,0 +1,328 @@
...
+        {
+          "properties": {
+            "side_id": {
+              "const": "runner"
+            },
+            "card_type_id": {
+              "enum": [
+                "event",
+                "hardware",
+                "runner_identity",
```
distributive, line 77: I'm glad we're separating these types, but this isn't consistent with the faction equivalent (`neutral-runner`, `neutral-corp`).

(Also note the hyphen/underscore discrepancy)

schema/v2/card_schema.json
```diff
@@ -0,0 +1,328 @@
...
+            "side_id": {
+              "const": "corp"
+            },
+            "faction_id": {
+              "enum": [
+                "haas-bioroid",
+                "jinteki",
+                "nbn",
+                "neutral-corp",
+                "weyland-consortium"
```
distributive, line 22: Should these use underscores instead of hyphens?

schema/v2/cycle_schema.json
```diff
@@ -14,18 +14,10 @@
...
-      "position": {
-        "minimum": 0,
-        "type": "integer"
-      },
-      "rotated": {
-        "type": "boolean"
-      },
-      "size": {
-        "minimum": 1,
-        "type": "integer"
```
distributive, line 26: What's the reasoning for scrapping this data?

v2/factions.json
```diff
@@ -1,84 +1,80 @@
 [
   {
     "code": "adam",
-    "color": "000000",
+    "description": "",
```
distributive, line 4: This field is missing from the other mini-factions entirely. Should it be present as an empty string for them too?

schema/v2/faction_schema.json
```diff
@@ -5,15 +5,54 @@
...
+    ],
     "properties": {
       "code": {
         "minLength": 1,
         "type": "string"
       },
-      "color": {
-        "description": "Faction color in hex triplet format. This can be used to display faction-specific information in the appropriate color on an HTML webpage.",
-        "maxLength": 6,
-        "minLength": 6,
```
distributive, line 58: I actually used these values as the embed color for Sahasrara's card outputs. Would it be possible to keep this information on NRDB?

distributive (commented):
Looks good! Just a few minor formatting things and some questions.

schema/v2/cycle_schema.json
distributive: Actually scratch that I get it.

schema/v2/card_schema.json
plural: what is inconsistent?  Is it just the underscores?  I was leaving off broader underscore-ification to it's own PR, but included this one as in scope for the verification in this schema.

schema/v2/card_schema.json
plural: leaving broader formatting changes to their own PR.

v2/factions.json
plural: the schema allows it to be present or absent. I don't personally have strong feelings about requiring everything to be present even if the default is empty, but let me know if you do.

schema/v2/faction_schema.json
plural: i have a preference for keeping color information and other formatting info outside.  I don't even like that we have HTML tags in the text and would prefer something like markdown to keep things a bit more format-flexible.

I am happy to throw these back in for now, though, if you like.

plural:
I added color back in.

v2/factions.json
distributive: I just noticed it was inconsistent. I don't mind either way.

schema/v2/card_schema.json
distributive: Corp/runner goes first in the identity enum but second in the faction enum. It's a super minor thing but something that's inevitably going to trip me up some day.

v2/factions.json
plural: ok, let's leave off optional fields instead of having them present with an empty value.

schema/v2/card_schema.json
plural: i didn't actually follow that, but i did sort the enums.

oh, are you talking about neutral-{corp,runner} vs. {corp,runner}_identity ?  I actually think that is natural.  We don't say "an identity that is a corp" we say "a corp id" and likewise, we don't say "a corp card that is neutral" but "a neutral corp card" typically.

Hopefully solid tests help prevent mixups thoroughly!

distributive (approved):

PR #666, opened 2022-06-20
V2 finish schema and testing first pass

plural:
Add schemas where they are missing, complete a first pass of tests and validation against the things the schema can't check for us.
Add test to validate text and stripped text between V1 and V2 files.


plural:
There is some tidying up that could be done here, but i am ok with doing that in a separate PR (for some of the reading of json files in nested trees in particular).

distributive (approved):

PR #671, opened 2022-06-21
Add faction colors for mini-factions.

plural:
Address #547 in this repo, at least.  NRDB ... doesn't appear to use these colors, but i know the discord bot does and we may use them in other places.

### DIFF
--- a/v2/factions.json
+++ b/v2/factions.json
@@ -78,7 +78,7 @@
   },
   {
     "code": "sunny-lebeau",
-    "color": "6E6E6E",
+    "color": "786196",
     "is_mini": true,
     "name": "Sunny Lebeau",
     "side_code": "runner"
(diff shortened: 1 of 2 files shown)


v2/factions.json
```diff
@@ -78,7 +78,7 @@
   },
   {
     "code": "sunny-lebeau",
-    "color": "000000",
+    "color": "6E6E6E",
```
distributive, line 81:
```suggestion
    "color": "786196",
```

distributive (changes requested):
They do actually appear on the site as profile colors:
image

Looking at this I think Sunny might be too close to the neutral color. I took a leaf from GLC's book and used the purple tone from her background:
image

plural:
my local instance is not showing anything different here.  :/

Additionally, the css has its own, literal, copy of the colors.  :/

distributive:
Classic NetrunnerDB

distributive (approved):

PR #672, opened 2022-06-22
Reformatting set names

distributive:
This unifies pack codes and cycle codes as being their full name in all lowercase with underscores instead of spaces (e.g. What Lies Ahead has gone from `wla` to `what_lies_ahead`)


plural:
I missed a test in Card Sets in validate_v2.ts:

+
+  it('has valid cycle ids', () => {
+    const cardCycleIds = new Set<string>(getCyclesV2Json().map(c => c.code));
+    sets.forEach(s => {
+      expect(cardCycleIds, `Card set ${s.name} has invalid card_cycle_id ${s.card_cycle_id}`).to.include(s.card_cycle_id);
+    });
+  });

the cycles.json file also needs to change "core" to "core_set" .

I found this when trying to import into the api serve.r

NoahTheDuke:
@plural I think your code formatting got converted to bullet points, lol

distributive:
I was able to copy it over still and it seems to work.

plural:
Nice, thank you!

PR #674, opened 2022-06-25
Illustrator consistency

plural:
tidies up illustrator fields.  Standardizes on comma-space separated for multiple credits.


pack/df.json
```diff
@@ -6,7 +6,7 @@
     "faction_code": "anarch",
     "faction_cost": 1,
     "flavor": "With each passing day alone, Hoshiko found it harder to think. With each hour, the static grew louder.",
-    "illustrator": "Photo Tammy Gann/Unsplash. Deep Dream",
+    "illustrator": "Tammy Gann/Unsplash, Deep Dream",
```
NoahTheDuke, line 9: Maybe reach out to Nisei creative for this one. The others make sense as either FFG or Nisei worked with both parties, but these deep dream ones might have different rules about attribution.

plural: yeah, i asked.  Folks weren't clear, so i'm changing this to be treated as a single illustrator.

NoahTheDuke (approved):

distributive (approved):
Do we want to consider having the illustrator field as a list rather than a comma-separated string or is that over complicating things?

plural:
yes to changing this to a list.  I'll do that separately, though.

PR #675, opened 2022-06-26
Renamed MWLs to Restrictions in v2

distributive:
We settled on "restriction" as a better term for ban lists, MWLs, points list, etc. This updates only the v2 json to reflect that.

Note that until my draft PR is ready this is a breaking change for the api server.


plural (approved):
This actually isn't a breaking change for the API server.  The mwl/restrictions aren't in use there yet.

PR #677, opened 2022-06-28
Add tests for ensuring consistency between v1/v2 cards & printings.

plural:
Additionally, validate card_set_id matches v2/printings/ file names.

### DIFF
--- a/pack/sg.json
+++ b/pack/sg.json
@@ -416,7 +416,7 @@
     "deck_limit": 3,
     "faction_code": "shaper",
     "faction_cost": 2,
-    "flavor": "\"You know there's no water in the Sea of Tranquility, right?\"\n\"That doesn't mean there's no beach.\"",
+    "flavor": "“You know there's no water in the Sea of Tranquility, right?”\n“That doesn't mean there's no beach.”",
     "illustrator": "Benjamin Giletti",
     "pack_code": "sg",
     "position": 21,
--- a/pack/sg.json
+++ b/pack/sg.json
@@ -416,7 +416,7 @@
     "deck_limit": 3,
     "faction_code": "shaper",
     "faction_cost": 2,
-    "flavor": "“You know there's no water in the Sea of Tranquility, right?”\n“That doesn't mean there's no beach.”",
+    "flavor": "\"You know there's no water in the Sea of Tranquility, right?\"\n\"That doesn't mean there's no beach.\"",
     "illustrator": "Benjamin Giletti",
     "pack_code": "sg",
     "position": 21,
--- a/src/index.ts
+++ b/src/index.ts
@@ -42,14 +42,14 @@ export function textToId(text: string): string {
     .join('_');
 }
 
-export const getCardSubtypesV2Json = makeV2Reader('subtypes');
 export const getCardSetsV2Json = makeV2Reader('printings');
+export const getCardSubtypesV2Json = makeV2Reader('subtypes');
 export const getCyclesJson = makeV1Reader('cycles');
 export const getCyclesV2Json = makeV2Reader('cycles');
-export const getPacksJson = makeV1Reader('packs');
 export const getFactionsJson = makeV1Reader('factions');
 export const getFactionsV2Json = makeV2Reader('factions');
 export const getMwlJson = makeV1Reader('mwl');
+export const getPacksJson = makeV1Reader('packs');
 export const getPrebuiltsJson = makeV1Reader('prebuilts');
 export const getRotationsJson = makeV1Reader('rotations');
 export const getSetTypesV2Json = makeV2Reader('set_types');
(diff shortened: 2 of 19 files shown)


pack/sg.json
```diff
@@ -416,7 +416,7 @@
     "deck_limit": 3,
     "faction_code": "shaper",
     "faction_cost": 2,
-    "flavor": "\"You know there's no water in the Sea of Tranquility, right?\"\n\"That doesn't mean there's no beach.\"",
+    "flavor": "“You know there's no water in the Sea of Tranquility, right?”\n“That doesn't mean there's no beach.”",
```
NoahTheDuke, line 419: Is this intentional?

src/index.ts
```diff
@@ -105,3 +105,21 @@ export function getCardsV2Json(): Record<string, any>[] {
   }
   return CARDS_JSON;
 }
+
+export function getPrintingsV2Json(): Record<string, any>[] {
+  const PRINTINGS_JSON: Record<string, any>[] = [];
+
+  if (PRINTINGS_JSON.length === 0) {
```
NoahTheDuke, line 112: This will always be true.

test/validate_v1_v2_equality.ts
```diff
@@ -1,36 +1,244 @@
...
   const v1Cards = getCardsJson();
-  const v1Text = new Map<string, string>();
-  const v1StrippedText = new Map<string, string>();
+  const v1CardsByTitle = new Map<string, any>();
   v1Cards.forEach(c => {
-    v1Text.set(c.title, c.text);
-    v1StrippedText.set(c.title, c.stripped_text);
+    v1CardsByTitle.set(c.title, c);
   });
+  const v2CardsByTitle = new Map<string, any>();
```
NoahTheDuke, line 11: Any particular reason for separate maps instead of a single map of id to card object?

test/validate_v1_v2_equality.ts
```diff
@@ -1,36 +1,244 @@
...
+
+  function validatev1V2Equality(fieldName, v1Getter, v2Getter) {
+    v1CardsByTitle.forEach((c, title) => {
+        expect(v1Getter(title), `${fieldName} mismatch for ${title}`).to.equal(v2Getter(title));
+    });
+  }
+
+  it('text matches', () => {
+    validatev1V2Equality('text',
+        function(title) { return v1CardsByTitle.get(title).text; },
```
NoahTheDuke, line 54: These could be arrow functions: `(title) => v1CardsByTitle.get(title).text`

Given that all of these are simply checking a property of an entry in the respective maps, the comparison function could take the two properties and check directly (written on a phone, apologies):
```
function checker(title, v1Prop, v2Prop) {
  expect(v1Map.get(title)[v1Prop]).to.equal(v2Map.get(title)[v2Prop]);
}
```

test/validate_v1_v2_equality.ts
```diff
@@ -1,36 +1,244 @@
...
+  });
+  const printings = getPrintingsV2Json();
+  const printingsById = new Map<string, any>();
+  printings.forEach(p => {
+    printingsById.set(p.id, p);
+  });
+
+  it('correct number of printings', () => {
+    v1CardsByCode.forEach((v, k) => {
+      expect(printingsById.has(k), `v1 code ${k} exists in printings id map`).to.be.true;
```
NoahTheDuke, line 169: Could also be `expect(print).to.have.keys(k)`

pack/sg.json
plural: yep.  the v2/printing version had fancy quotes.

src/index.ts
plural: yeah, leftover from a misguided caching attempt when i first started this file.  Removed.

pack/sg.json
distributive: As far as I can tell there's no difference between the quote marks on VRcation and other flavour text in Gateway. Should v2 be changed to be consistent rather than this changed to match v2?

test/validate_v1_v2_equality.ts
plural: ooh, nice.  That tidies things up quite a bit.   Thanks for the idiomatic JS/TS suggestions!

pack/sg.json
NoahTheDuke: No other card with quotes uses fancy quotes. I don't think we should be using them here either.

NoahTheDuke: Maybe I misunderstood your comment that you've reverted this change, in which case my apologies.

test/validate_v1_v2_equality.ts
plural: there is something funny happening with the typescript map interaction.  That doesn't work here.  Same with using a Set().

plural:
Thanks!

test/validate_v1_v2_equality.ts
plural: yeah, that doesn't work either.  it actually makes the test timeout.  🤷

NoahTheDuke:
Sorry I keep noticing things every time I look instead of all at once. I know how annoying it is when it's happened to me, lol. Thanks so much for writing all of this, it's really nice to have.

plural:
hehehe, no worries, Noah.  Thanks for the careful look and the suggestions.  This is much neater now.

test/validate_v1_v2_equality.ts
```diff
@@ -1,36 +1,188 @@
...
+      const v2Subtypes = v2.subtypes ? v2.subtypes : [];
+      expect(v1Keywords, `subtypes mismatch for ${title}`).to.deep.equal(v2Subtypes);
+    });
+  });
+
+  function validatev1V2Equality(fieldName, v1Field, v2Field) {
+    v1CardsByTitle.forEach((c, title) => {
+      expect(v1CardsByTitle.get(title)[v1Field], `${v2Field} mismatch for ${title}`).to.equal(v2CardsByTitle.get(title)[v2Field]);
+    });
+  }
```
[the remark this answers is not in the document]
plural: it wasn't happy with the undef check as you had it, but making a local variable with a check against the param did the trick.

NoahTheDuke (approved):

PR #680, opened 2022-06-30
Ids and updated object names

plural:
Updating all the JSON file and schema names for v2 to use the new names.

Updated remain objects to use id instead of code where appropriate.  Added missing _id values.


plural:
i will leave this as a draft until distributive's API changes are in and then i'll handle the udpates on the API server side from this.

distributive (approved):

PR #681, opened 2022-06-30
Updates to card pools and snapshots

distributive:
- Card pools now specify the ID of the singular format they belong to
- Format snapshots are now given IDs within the json


v2/card_pools/standard.json
```diff
@@ -2,6 +2,7 @@
   {
     "code": "pre_rotation",
     "name": "Pre Rotation",
+    "format_id": "standard",
```
plural, line 5: would you please add a test that ensures that all the format_ids in these files match the file names?

plural (changes requested):

v2/card_pools/standard.json
distributive: Done!

plural (approved):

PR #691, opened 2022-07-14
Set up banlist 22.08 and rotation.

plural:


v2/restrictions/standard/standard_ban_list_22_08.json
```diff
@@ -0,0 +1,88 @@
+{
+  "id": "standard_ban_list_22_08",
+  "name": "Standard Ban List 22.08",
+  "date_start": "2022-08-05",
+  "format_id": "standard",
+  "subtypes": {
+    "banned": ["current"]
+  },
+  "banned": [
+    "cyberdex_sandbox",
```
NoahTheDuke, line 10: Should these be sorted?

plural: none of them are sorted really at the moment.

I looked quickly at adding the sorting plugin to prettier, but it messed with the subtypes arrays for cards, which are sorted by choice, not alphabetically.  

i can add a test to sort the values just for these, though.  It would be great to have and keep them sorted.

NoahTheDuke: I had a feeling it would be complicated, so no worries.

plural: not terribly complicated.  :D

NoahTheDuke (approved):

PR #694, opened 2022-07-15
Midnight Sun spoiler season day 4 cards!

plural:


pack/ms.json
```diff
@@ -321,6 +321,27 @@
...
+    "flavor": "At depths no human tech can reach, a Trieste proxy can manipulate a mindscape with unparalleled precision.",
+    "illustrator": "Dimik",
+    "keywords": "Bioroid",
+    "pack_code": "ms",
+    "position": 34,
+    "quantity": 3,
+    "side_code": "corp",
+    "stripped_text": "When you rez this asset, choose 1 rezzed piece of bioroid ice. Runner card abilities cannot break subroutines on the chosen ice. Designed by 2019 NISEI World Champion Oliver \"Pinsel\" Siccha",
+    "stripped_title": "Trieste Model Bioroids",
+    "text": "When you rez this asset, choose 1 rezzed piece of <strong>bioroid</strong> ice.\nRunner card abilities cannot break subroutines on the chosen ice.\n<strong>Designed by 2019 NISEI World Champion Oliver \"Pinsel\" Siccha</strong>",
```
distributive, line 339: Until now, the "designed by" text has always been in the flavour text. I've just realised this is the first champion card to have regular flavour text alongside that so maybe we should change how we represent that. For now should it be moved back to flavour?

distributive (changes requested):

pack/ms.json
plural: ooh!  Thanks for catching this.  I updated the converter and fixed this one.

distributive (approved):

PR #697, opened 2022-07-16
Midnight Sun spoiler season day 5 cards!

plural:


pack/ms.json
```diff
@@ -342,6 +383,27 @@
     "type_code": "asset",
     "uniqueness": false
   },
+  {
+    "code": "33035",
+    "cost": 2,
+    "deck_limit": 3,
+    "faction_code": "haas-bioroid",
+    "faction_cost": 2,
+    "flavor": "End the run. End the run. End the run. End the run.",
```
distributive, line 392: Is there any way to format this like on the card? (Getting incrementally smaller)

I imagine not especially if we're going to move to markdown at some point.

distributive (approved):

pack/ms.json
plural: yeah, probably not doable in a durable way right now.  We could look more closely at formatting for text and flavor text as we get the new site up and running, though!

PR #702, opened 2022-07-19
Midnight Sun spoiler season day 8 cards!

plural:


schema/v2/cards_schema.json
```diff
@@ -198,8 +205,15 @@
...
-      "type": "integer"
+      "oneOf": [
+        {
+          "minimum": 0,
+          "type": "integer"
+        },
+        {
+          "type": "null"
+        }
+      ]
```
distributive, line 216: I don't know how I feel about this. Until a card is printed with multiple variables on it this technically works, but I don't like that the true meaning of "null" here is implicit.

distributive (approved):

schema/v2/cards_schema.json
plural: i'm not sure i follow what you mean by multiple variables.  This is what we do for fields with nullable values where we otherwise need a constraint if there is a value.  I can look a bit more at the schema spec but this is doing our "if null or an integer this is a fine value" check.

plural: i wonder if we should make X explicit instead of null.   Is that what you were getting at, @distributive ?

PR #707, opened 2022-07-23
Tidy up champ formatting.

plural:


distributive:
Is the plan to make all champion texts explicitly bold? Also is this how we want to format Echo's flavour text? Technically none of it's bold.

pack/ms.json
```diff
@@ -672,7 +672,7 @@
     "deck_limit": 3,
     "faction_code": "haas-bioroid",
     "faction_cost": 2,
-    "flavor": "At depths no human tech can reach, a Trieste proxy can manipulate a mindscape with unparalleled precision.\n<champion>Designed by 2019 NISEI World Champion Oliver \"Pinsel\" Siccha</champion>",
+    "flavor": "At depths no human tech can reach, a Trieste proxy can manipulate a mindscape with unparalleled precision.\n<champion><strong>Designed by 2019 NISEI World Champion Oliver \"Pinsel\" Siccha</strong></champion>",
```
NoahTheDuke, line 675: I thought `<strong>` was implied by `<champion>`.

pack/ms.json
```diff
@@ -693,7 +693,7 @@
     "deck_limit": 3,
     "faction_code": "haas-bioroid",
     "faction_cost": 2,
-    "flavor": "End the run. End the run. End the run. End the run.",
+    "flavor": "<strong>End the run.</strong> <strong>End the run.</strong> <strong>End the run.</strong> <strong>End the run.</strong>",
```
NoahTheDuke, line 696: These are the ones that get smaller and smaller, right? Is `<strong>` appropriate?

NoahTheDuke:
Ope, @distributive already said both of my comments lol.

pack/ms.json
plural: in the current NRDB css, but that's it.  Also, the MS sheet has the bold tags, so the converter brings them through.  I'm inclined to keep this to avoid manual exceptions, but it's worth revisiting the CSS and tidying up the conventions in the dev sheets.

pack/ms.json
plural: the dev sheet has these as {b1}End the run.{/b1} {b2}End the run.{/b2} ... {/b4}.

Ignoring them is the other option, but mapping them to a distinct intent is also doable (with corresponding CSS changes).

Again, i dislike the hodgepodge pile of formatting junk.  bleargh.

pack/ms.json
NoahTheDuke: Should we remove the champion tag entirely then? Feels like we're double specifying.

pack/ms.json
NoahTheDuke: In my opinion, I feel like we shouldn't include them. Software relying on this will produce incorrect styling unless they special case this one, and it feels awkward to add new tags specifically for this card.

distributive: There is some precedence for resized flavour text:

NBN: The World is Yours

plural: we don't do anything different for that flavor text for The World Is Yours in the JSON, though.

distributive: I think we should either add formatting to replicate the layout of the text on the card (that tools can ignore if its incompatible with them) or not do anything at all. Making the text on Echo bold will just confuse people for the benefit of matching a private document.

plural: okee dokee, taking the formatting off of Echo.

plural:
ok, I tidied up my tidying up.  :D

<champion> -> <strong> simply.

Reverted Echo.

NoahTheDuke (approved):
I can work with this!

PR #709, opened 2022-07-29
Update card pools and formats for midnight sun.

plural:


v2/card_pools/startup.json
```diff
@@ -16,5 +16,17 @@
       "system_gateway",
       "system_update_2021"
     ]
+  },
+  {
+    "id": "startup_midnight_sun",
```
distributive, line 21: Do we want to call it this? We essentially now have three Startups per cycle (just that cycle; that cycle and the next booster pack; that cycle and the first pack of the next cycle).

My suggestion might be:
- Startup (Cycle)
- Startup (Cycle)+
- Startup (Cycle+Next)

distributive (approved):

PR #712, opened 2022-08-19
Update illustrator credits for NtscapeNavigator.

plural:
at NtscapeNavigator's request.


NoahTheDuke:
On the one hand, I want to respect @nttscapenavigator's request. On the other hand, isn't the illustrator field meant to represent what is printed?

plural:
the remaster will be updating credits and we did similar things for Jess Horig's champ card (not the illustrator field, but the same principle) and seems like the obviously correct and kind thing to do.

NoahTheDuke:
Cool.

distributive (approved):

PR #714, opened 2022-08-31
Add legacy_code fields to Card Cycles and Sets.

plural:
This enables interop with NRDB classic.  We had printing codes, but not legacy codes for card sets and card cycles.


NoahTheDuke (approved):
Looks good to me.

PR #715, opened 2022-08-31
Fix cycle and set specification for startup card pools.

plural:
We had a slightly wonky specification for the card pools here.

I'll add a test as a followup, but card pools should list cycles for any cycles that are completely encompassed and the sets should also be spelled out explicitly.

It's OK to have 1 set from a cycle specified, but not to have a missing set from a cycle for eternal, standard, snapshot, or startup.

### DIFF
--- a/v2/card_pools/startup.json
+++ b/v2/card_pools/startup.json
@@ -3,30 +3,44 @@
     "id": "startup_ashes",
     "format_id": "startup",
     "name": "Startup (Ashes)",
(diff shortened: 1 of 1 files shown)


NoahTheDuke (approved):
i trust this is correct, idk enough about start up

plural:
hehehe.  i went back to the original announcement to make sure this was
correct.  :)

-jason

PR #716, opened 2022-08-31
Card pools have all sets for cycles

plural:


plural:
i thought this was annoying until i wanted to write a test for it.

I think this is better because:
- an explicit "is this cycle legal?" check is easier (can be done w/just this data).
- an explicit "is this set legal?" check is easier (can be done w/just this data).
- prevents typos and gaps for any sets.
- still allows stray sets (but not incomplete cycles if specified in card_cycle_ids).

distributive (approved):

PR #717, opened 2022-09-06
Cleaned up card set types

distributive:
- Added `booster_pack` as a set type to represent the new trend of releasing booster packs next to full releases
- Made Salvaged Memories an expansion
- Made the constructed Terminal Directive set a deluxe

My reasoning for changing SM and TD from being data packs is it feels inaccurate to call a set that wasn't part of a cycle a data pack. I'm happy to revert these if they make more sense as data packs.

Here's how I'm viewing each type:
- booster_pack (a booster pack; so far a direct subset of another set)
- campaign (Terminal Directive Campaign)
- core (the core sets and system update)
- data_pack (the major releases of each cycle; 6 per FFG cycle and 2 per NISEI cycle)
- deluxe (a set released in a big box)
- draft (draft)
- expansion (miscellaneous Standard set)
- promo (official non-Standard set; NAPD multiplayer)

Would it be useful to add a description field to `card_set_types.json` explaining what all the types represent?

### DIFF
--- a/schema/v2/card_set_types_schema.json
+++ b/schema/v2/card_set_types_schema.json
@@ -15,8 +15,13 @@
         "description": "Name of the type of set.",
         "minLength": 1,
         "type": "string"
+      },
+      "description": {
+        "description": "A brief explanation of the type of set.",
+        "minLength": 1,
+        "type": "string"
       }
     },
-    "required": ["id", "name"]
+    "required": ["id", "name", "description"]
   }
 }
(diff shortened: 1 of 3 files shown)


NoahTheDuke:
Thanks for doing all this.

I think adding a description would be good. I created the initial set of types without too much thought and never revisited them so I'm glad you're cleaning them up.

NoahTheDuke (approved):

plural:
this is great.  Would you please follow up and add the descriptions to the api server app as well?

plural (approved):

PR #719, opened 2022-09-17
remove rotated cards from SBL 22.08 and SBL 22.09

francescopellegrini:
Might as well slim them a bit


distributive (changes requested):
Thanks for this! I totally forgot
It looks like Employee Strike accidentally got removed. Also, would you be able to update the v2 data as well? (see `v2/restrictions/standard/standard_ban_list_22_09.json` and the `22-08` equivalent)

mwl.json
```diff
@@ -2989,33 +2938,6 @@
       "09023": {
         "deck_limit": 0
       },
-      "09053": {
-        "deck_limit": 0
-      },
```
distributive, line 2994: This is Employee Strike (Data and Destiny) which hasn't rotated

mwl.json
```diff
@@ -3241,33 +3112,6 @@
...
```
distributive, line 3246: Same here

distributive (changes requested):
Forgot to add these to the review

francescopellegrini:
> Thanks for this! I totally forgot It looks like Employee Strike accidentally got removed.

@distributive thank you so much for pointing it out!

distributive (approved):
This looks good to me!

plural (approved):

PR #720, opened 2022-09-20
Added v1/v2 mwl consistency tests

distributive:
This tests:
- are there the same number of mwl entries in both versions
- do the entries share the same names
- do the entries have matching data on bans, restricted cards, universal faction costs, and global penalties

I'm not convinced this is the best, fastest, or cleanest way to do this, but it seems to have caught most things. It doesn't test for points since they were never used in standard.


test/validate_v1_v2_equality.ts
```diff
@@ -248,4 +259,108 @@ describe('Printings v1/v2 equality', () => {
...
+  getCardsV2Json ().forEach(c => {
+    printingsByCardId.set(c.id, []);
+  });
+  printings.forEach(p => {
+    printingsByCardId.get(p.card_id).push(p);
+  });
+
+  function v1ToBanned(mwl) {
+    const arr: [string, Record<string, number>][] = Object.entries(mwl.cards);
+    return arr.filter(card => card[1].deck_limit != undefined && card[1].deck_limit == 0)
```
plural, line 284: nit, but you don't need the undefined check.

test/validate_v1_v2_equality.ts
```diff
@@ -248,4 +259,108 @@ describe('Printings v1/v2 equality', () => {
...
+      if (!obj[card[1].global_penalty]) {
+        obj[card[1].global_penalty] = [];
+      }
+      obj[card[1].global_penalty].push(card[0]);
+    });
+    return obj;
+  }
+
+  function cardsToPrintings(cards) {
+    if (!cards)
```
plural, line 316: style thing for me, but i dislike conditionals with no braces.  this isn't incorrect, obviously.

test/validate_v1_v2_equality.ts
```diff
@@ -248,4 +259,108 @@ describe('Printings v1/v2 equality', () => {
...
-});
\ No newline at end of file
+});
+
+describe('MWLs v1/v2', () => {
+  const standardRestrictions = getRestrictionsV2Json()['standard'];
+  const mwls = getMwlJson();
+  const pairs = mwls.map(mwl => [mwl, standardRestrictions.find(r => r.name == mwl.name)]);
+
+  const printings = getPrintingsV2Json ();
```
plural, line 269: remove space after method name.

test/validate_v1_v2_equality.ts
```diff
@@ -248,4 +259,108 @@ describe('Printings v1/v2 equality', () => {
...
+  printings.forEach(p => {
+    printingsByCardId.get(p.card_id).push(p);
+  });
+
+  function v1ToBanned(mwl) {
+    const arr: [string, Record<string, number>][] = Object.entries(mwl.cards);
+    return arr.filter(card => card[1].deck_limit != undefined && card[1].deck_limit == 0)
+              .map(card => card[0]).sort();
+  }
+  function v1ToRestricted(mwl) {
```
plural, line 287: all of the prep functions here would benefit from some comments.  I had to read a couple of times to clearly picture what the output was here.

plural (approved):
couple of style and doc things, but thank you for adding these tests!

test/validate_v1_v2_equality.ts
NoahTheDuke: If you're ever worried about undefined like this, you can say `card[1]?.deck_limit === 0`, which is equivalent to the undefined check but much shorter.

PR #727, opened 2022-10-17
Moved card design attributions to a separate field to flavor

distributive:
In this version "attribution" is a property of printings, however it could be argued it should be a property of cards themselves.

I'm happy to redo with that change if requested.


plural:
I do think this is better as a function of the card.

Additionally, this will remove the attribution from NRDB once pushed.  It's annoying, but maybe the old files should stay as is and the v1/v2 tests should be updated to add the attribution to the flavor text for a comparison to keep things in sync still?

plural (approved):
Looks great. Thank you.

PR #728, opened 2022-10-25
update text

mankinchi:
This is to handle the issue mentioned here https://github.com/NetrunnerDB/netrunnerdb/issues/623

This is to work with the new way to render Trace tag in https://github.com/NetrunnerDB/netrunnerdb/pull/705


NoahTheDuke:
I don't think it's a good idea to remove the word "trace" from the text. This should be a repository of the text that is printed on the cards; that we style it in different ways is purely a UI issue.

NoahTheDuke (changes requested):
As above.

plural:
hiya!  Would you still like to proceed with this?  If so, let us know what you need.  Thanks.  :)

mankinchi:
Hi,

Sorry for the delay. I have been caught up with stuff in life and unable to work on this for awhile. If you want, feel free to abandon this and move forward. I think I will look back into this in 2-3 weeks time

Kind regards,
*
*
*Tri Nguyen*

IT Consultant 

M: 0449 258 526

mankinchi:
so sorry for the disappearance. Personal issue is annoying but resolved. I will look back into this and update it asap 😄

plural:
no worries at all!

plural:
Would you still like to try to do this or have one of us take care of it?

plural:
i ended up taking care of this in #742 .  Thanks for kicking this off, though!

PR #732, opened 2022-12-02
Parhelion Scoops Drop 1!!!!

plural:


v2/card_pools/startup.json
```diff
@@ -30,12 +30,7 @@
     "id": "startup_ashes_plus_midnight_sun",
     "format_id": "startup",
     "name": "Startup (Ashes + Midnight Sun)",
-    "card_cycle_ids": [
-      "ashes",
-      "borealis",
-      "system_gateway",
-      "system_update_2021"
-    ],
+    "card_cycle_ids": ["ashes", "system_gateway", "system_update_2021"],
```
distributive, line 33: This shouldn't change, unless we want the intermediate versions of Startup between full releases (i.e. the versions of Startup containing 1.5 cycles) to be represented temporarily.

v2/card_pools/startup.json
```diff
@@ -45,5 +40,18 @@
...
     ]
+  },
+  {
+    "id": "startup_borealis",
+    "format_id": "startup",
+    "name": "Startup (Borealis)",
+    "card_cycle_ids": ["borealis", "system_gateway", "system_update_2021"],
+    "card_set_ids": [
+      "midnight_sun",
+      "midnight_sun_booster_pack",
```
distributive, line 51: Minor point, it's not actually important either way, but do we want to include the booster packs in set lists where those packs are strict subsets of other sets in the list? In this case the set of cards doesn't change if we remove MSBP from the list.

packs.json
```diff
@@ -386,6 +386,15 @@
     "position": 1,
     "size": 20
   },
+  {
+    "code": "ph",
+    "cycle_code": "borealis",
+    "date_release": "2022-12-12",
```
distributive, line 392: Should this be the 12th or the 9th?

distributive (changes requested):

v2/card_pools/startup.json
plural: right now we have a test that if a cycle is specified, all packs in the cycle are specified to make sure we don't miss things.  we can discuss revisiting that, but i don't wanna do that in this PR.

v2/card_pools/startup.json
plural: this is the same issue as the other comment.  If we kept borealis, we would be *forced* to have parhelion in here as well.  Removing borealis lets us keep just the one set in card_set_ids without any failing tests.

packs.json
plural: confirmed we are going with the digital release date of the 9th. thanks for flagging!

distributive (approved):

PR #742, opened 2022-12-24
Remove trace and normalize trace[X|#] use.

plural:


pack/23s.json
```diff
@@ -130,9 +130,9 @@
     "position": 7,
     "quantity": 3,
     "side_code": "runner",
-    "stripped_text": "This card is not trashed until another current is played or an agenda is scored. Whenever you steal an agenda, force the Corp to \"Trace 0 If unsuccessful, the Runner gains credits equal to the number of agenda points in both players' score areas.\"",
+    "stripped_text": "This card is not trashed until another current is played or an agenda is scored. Whenever you steal an agenda, force the Corp to \"Trace[0]. If unsuccessful, the Runner gains credits equal to the number of agenda points in both players' score areas.\"",
```
distributive, line 133: In this stripped text the trace is formatted with square brackets.

pack/23s.json
```diff
@@ -314,7 +314,7 @@
     "side_code": "corp",
     "stripped_text": "After you resolve this operation, your action phase ends. Play only if the Runner made a run during their last turn. Trace 4. If successful, give the Runner 4 tags.",
```
distributive, line 315: In this stripped text the trace is not formatted with brackets.

distributive (changes requested):
The formatting of traces in `stripped_text` fields seems inconsistent.

pack/23s.json
plural: yeah, that's intentional.  It should be the case throughout.  this enables x:"trace[3]" searches to do the right thing.

pack/23s.json
plural: *this one* was not intentional.  :)

distributive (approved):
I can't find any more!

PR #743, opened 2023-01-01
Removed "printed" fields from v2 printings

### LINKED ISSUE #669: Do the v2 printing cards need text fields?
We currently have printed_text and stripped_printed_text, but do we need it?  We don't want printings to have different text and have already taken an "update everything at once" approach when making text and rules changes.

Do we need these fields at all in v2?

plural:
Fixes #669

### DIFF
--- a/schema/v2/printings_schema.json
+++ b/schema/v2/printings_schema.json
@@ -33,15 +33,9 @@
       "printed_is_unique": {
         "type": "boolean"
       },
-      "printed_text": {
-        "type": "string"
-      },
       "quantity": {
         "minimum": 1,
         "type": "integer"
-      },
-      "stripped_printed_text": {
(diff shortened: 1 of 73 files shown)


NoahTheDuke:
Was there internal discussion about this? What led to the change?

plural:
yeah, we talked about it on discord a bit.  We aren't using these fields and if we want to really track historical text changes, it's a bit different than what this was doing.

distributive:
Do we want to give the same treatment to the `printed_is_unique` field?

plural:
Yes, i looked and we are currently enforcing that printed_is_unique is the same as is_unique in v1 so i went ahead and removed that here as well.

distributive (approved):

PR #746, opened 2023-01-12
Multi-sided cards

distributive:
This is a WIP attempt to formalise double-faced cards (e.g. Hoshiko) and cards with functional variants (e.g. Jinteki Biotech).

My approach has been to give cards and printings an optional `"sides"` field (working name) that defines a series of objects representing a subset of fields of their parent. This allows the sides to only redefine properties that are different, while leaving the rest implicitly the same:

```json
{
  {
    "title": "Card Name",
    "text": "Flip this card to see side 2.",
    "sides": [
      {
        "text": "Flip this card to see side 1."
      }
    ]
}
```

The downside of this is it's a bit asymmetric and can't always accurately represent what the flip side/other variants represent.

For instance, each copy of Matryoshka has its own flavour text, but none of them (printing order aside) takes precedence over the others, even though this implementation requires that. Also, while flip cards are trivial to understand (one side is always the initial side) Jinteki Biotech is actually three double-sided cards, represented here as 4 objects (the parent card and its three sides). 

Applying this system to the Terminal Directive Campaign cards has similar issues, and rewriting the validation tests is going to be a real hassle.

An alternative system might be to represent the sides as separate cards/printings and have them reference each other in some way (this is effectively how it was implemented for TDC). This makes them easier to search for and avoid some of the above issues, but we'd need a way to coalesce cards/printings that represent the same card.


NoahTheDuke:
Scryfall splits multiface cards into different "faces" that are referenced in a card_faces array on the base card object, and has a "layout" attribute that arranges the different faces. If a given card is "normal", then instead of card_faces, the single face info is merged into the base object.

I don't think magic has cards like Biotech where the same bad card has multiple different flipped sides, but I don't think that fundamental invalidates the model.

Think that would work for nrdb?

NoahTheDuke:
Examples from Scryfall/Magic the Gathering (make sure to click through to "Copy-pasteable JSON"):

* [Double-faced, with hint](https://scryfall.com/card/khm/A-40/a-alrund-god-of-the-cosmos-a-hakka-whispering-raven)
* [Double-faced (transform)](https://scryfall.com/card/soi/49/aberrant-researcher-perfected-form)
* [Flip (top and bottom)](https://scryfall.com/card/chk/153/akki-lavarunner-tok-tok-volcano-born)
* [Split (side by side)](https://scryfall.com/card/dgm/121/alive-well)

Here's the full list: [Scryfall layouts](https://scryfall.com/docs/api/layouts)

distributive:
Magic is weird.

distributive:
That's handy though. It looks like they do it broadly like I've done. Does Magic have an equivalent to Jinteki Biotech or a TDC card like Machilocation ([A](https://netrunnerdb.com/en/card/14010) and [B](https://netrunnerdb.com/en/card/14011))

NoahTheDuke:
No worries. Glad I could help.

NoahTheDuke:
Something to be explicit about, I think Scryfall's faces are not just properties that are different from the base card. I think they're intended to be "full" information that is merged into the base card object.

From my reading of the json, the double face cards don't have a base "text" property, each face has that. I think the first face entry in the array is considered the "upright" or "front", depending on the style, but that is less meaningful with the split cards (which we don't have to worry about).

I don't want to saddle you with a bunch of work lol, so please ignore any and all advice I give if you think it's too much work.

That said, I think it might be smart to determine a set of card data that can or could change depending on face, maybe treating each like a printing? And then put the full info into the face entries.

Having individual card entries for each side would complicate our data model, but would allow us to accurately handle biotech and matrioshka and sync (which I think has the same name on both sides).

lol This is a classic case of edge case poisoning. So annoying.

NoahTheDuke:
Your layouts have nice names. 👍

distributive:
The MtG system makes this so much easier to work with. I think it lets us deal with all the flip cards, the campaign cards, and Biotech trivially.

_Matryoshka_ on the other hand...

distributive:
The problem with Matryoshka is as far as the card is concerned it's a regular card, but the printing itself has multiple versions.

I don't want the card object to know it has differing flavour text, but shoehorning the layout system onto printings feels weird to me.

v2/layout.json
```diff
@@ -0,0 +1,27 @@
+[
+  {
+    "id": "normal",
+    "description": "A regular one-sided card.",
+    "name": "Normal"
+  },
+  {
+    "id": "flip",
+    "description": "A card with a front and back side.",
+    "name": "Flip"
```
distributive, line 10: e.g. Hoshiko

v2/layout.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  {
+    "id": "copy",
+    "description": "A card that is made up of multiple distinct cards with the same name.",
+    "name": "Copy"
```
distributive, line 15: e.g. [Machilocation][(url)](https://netrunnerdb.com/en/card/14010) and other campaign cards

v2/layout.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  {
+    "id": "facade",
+    "description": "A card that is made up of multiple cards with the same front face, but distinct flip sides.",
+    "name": "Facade"
```
distributive, line 20: Only Biotech

v2/layout.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  {
+    "id": "progression",
+    "description": "A card with iterations on its values.",
+    "name": "Progression"
```
distributive, line 25: e.g. [Evidence Collection](https://netrunnerdb.com/en/card/14000) and other campaign cards with the sticker mechanic

schema/v2/cards_schema.json
```diff
@@ -265,6 +265,17 @@
         }
       ]
     },
+    "layout_id": {
+      "oneOf": [
+        {
+          "type": "string",
+          "enum": ["normal", "flip", "copy", "facade", "progression"]
```
plural, line 272: single_sided instead of normal?  Not convinced it's better, but wdyt?

test/validate_v2.ts
```diff
@@ -233,6 +241,15 @@ describe('Printings', () => {
       });
     });
   });
+
+  it('printing files have valid copy_quantity attributes', () => {
+    printingsByFilename.forEach((printing, file) => {
+      const cardSetIdFromFile = basename(file).replace('.json', '');
```
plural, line 247: please remove unused variable.

v2/printings/parhelion.json
```diff
@@ -250,11 +250,35 @@
...
     "card_set_id": "parhelion",
-    "flavor": "{Card 6} Look at the little baby, even she helps!\n{Card 5} Sheʼs little—but fierce.\n{Card 4} Just as brave as his sisters.\n{Card 3} Sheʼs a typical middle child, really.\n{Card 2} Always there to back up her big sister.\n{Card 1} She leads and they all follow.",
+    "flavor": "She leads and they all follow.",
     "id": "33094",
     "illustrator": "Ed Mattinian",
     "position": 94,
-    "quantity": 6
+    "quantity": 6,
+    "layout_id": "copy",
+    "copy_quantity": 1,
```
plural, line 259: how about dropping this field out here, but keeping "quantity" in the sides?  You can still sum the quantities up for the test, but move *all* instances of the field into sides where they vary, rather than treating one differently.

wdyt?

We are going to need to do some formatting shenanigans for these cards either way.

test/validate_v1_v2_equality.ts
```diff
@@ -211,10 +273,17 @@ describe('Printings v1/v2 equality', () => {
...
     const v2Field = maybeV2Field == undefined ? v1Field : maybeV2Field;
     v1CardsByCode.forEach((c, code) => {
+      if (checkTdc(c, code)) {
+        return;
+      }
       expect(c[v1Field], `${v2Field} mismatch for ${code}`).to.equal(printingsById.get(code)[v2Field]);
     });
   }
 
+  function checkTdc(v1: any, code: string): boolean {
```
plural, line 283: you have this function defined twice, but with different signatures.  Which do you need?

v2/cards/bmi_buffer.json
```diff
@@ -10,5 +10,12 @@
   "stripped_text": "Whenever a program is trashed from your grip, host it on BMI Buffer instead of adding it to your heap. click click: Install 1 hosted program (paying all costs).",
   "stripped_title": "BMI Buffer",
   "text": "Whenever a program is trashed from your grip, host it on BMI Buffer instead of adding it to your heap.\n[click][click]: Install 1 hosted program (paying all costs).",
-  "title": "BMI Buffer"
+  "title": "BMI Buffer",
+  "layout_id": "progression",
+  "sides": [
```
plural, line 15: i'm torn on having the default for cards like this having some parts outside of the sides and some parts inside.

v2/cards/jinteki_biotech_life_imagined.json
```diff
@@ -7,9 +7,33 @@
...
   "stripped_title": "Jinteki Biotech: Life Imagined",
   "subtypes": ["division"],
-  "text": "Before taking your first turn, you may switch this identity with any copy of Jinteki Biotech.\n<strong>[click][click][click]:</strong> Flip this identity.\nThe Brewery: When you flip this identity, do 2 net damage.\nThe Tank: When you flip this identity, shuffle all cards in Archives into R&D.\nThe Greenhouse: When you flip this identity, place 4 advancement counters on 1 installed card that you can advance.",
-  "title": "Jinteki Biotech: Life Imagined"
+  "text": "Before taking your first turn, you may switch this identity with any copy of Jinteki Biotech.\n<strong>[click][click][click]:</strong> Flip this identity.",
+  "title": "Jinteki Biotech: Life Imagined",
+  "layout_id": "facade",
+  "sides": [
+    {
+      "stripped_text": "When you flip this identity, do 2 net damage.",
```
plural, line 18: This is a great example of how elegant this is!

plural (commented):
First, thank you for tackling this!

A few thoughts.

* TD Campaign cards:  I'm not sure we should care about them.  They aren't now and don't plan on being legal in any formats, they are mostly here as a historical record.  Should we do anything to them at all?  I'd give some thought to not doing anything with them.
* V1 cards.  I would also consider doing the minimal change here to keep existing NRDB useful and treating V1/V2 differences for just these cards in the test files.   I am certainly not against changing V1, but not losing info in the current system is probably useful.  That being said, this is a mild take.  :)

I have one open design question for you, but this is overall great!

NoahTheDuke:
Idk how much weight my voice has, but I think y'all should support the campaign cards. They don't require much to track and they're interesting and mildly relevant. I'm a maximalist in this respect.

v2/cards/bmi_buffer.json
distributive: Part of why I did it this way was to reduce the need to acknowledge extra sides when reading from the json. If, for some reason, we don't then there's still a complete card defined outside of the sides.

It also reduces the duplicate data (since the base card defines everything and the sides overwrite only what changes. It also means the overall structure of the `sides` attribute doesn't change between layout types.

I'm not actually wholly against separating the base version into its own side, but I can't think of any reason why that would allow us to define something we can't already, and once it's in the API we can format it however we want.

v2/printings/parhelion.json
distributive: Technically the base card is also a copy. I know it's sort of weird to have 6 equal cards have one arbitrarily denoted to dominant version, but see my comment below.

distributive:
I'd prefer to keep the campaign cards in as well, at least in part because the work for it is now done and "we might as well". One of the layout types (progression) becomes unused without the set.

I also think despite NSG not having anything to do with the cards, we can't assume nobody wants that data.

distributive:
I think the only changed I made to v1 were in the flavor text and Biotech, which I just changed to make the validation convenient. 

I can change the standard for flavour back to "\nFlip side:\n" (although I'd still like to make the ft on Matryoshka a bit more aesthetic).

Biotech we can just ignore if we really want, but I made the change since there's no way to validate the existing text without hardcoding something.

schema/v2/cards_schema.json
distributive: I have no preference. On a technicality not all non-normal layouts are multi-sided (copy cards are distinct single-sided cards, and the only printed progression cards use stickers). I'm not going to argue for either though; I'll make it whatever you think is best.

test/validate_v1_v2_equality.ts
distributive: They're in separate blocks:
- `describe('Card Sets v1/v2', () => {`
- `describe('Printings v1/v2 equality', () => {`

schema/v2/cards_schema.json
NoahTheDuke: I think "single_sided" is more descriptive than "normal".

v2/cards/bmi_buffer.json
plural: yeah.  I think "net less work overall" is the right way to resolve this.  :)

Handling the schema files to support "empty fields unless they are specified in every side" is ... gonna be super annoying.

Also, you reminded me that this does minimize work on existing readers, so huzzah!

test/validate_v2.ts
```diff
@@ -244,12 +248,23 @@ describe('Printings', () => {
 
   it('printing files have valid copy_quantity attributes', () => {
     printingsByFilename.forEach((printing, file) => {
```
plural, line 250: here and on 259, file is unused.  

nit: i wish github added a "Please fix" button here instead of a link to the action that found that.

plural (changes requested):
there is also still one warning for 'code' being an unused parameter as well.

schema/v2/cards_schema.json
distributive: I'm happy to concede to majority opinion.

distributive:
As per our last meeting the plan is to accept the schema for this will be a bit janky and put an issue up to eventually revisit this. If someone can have another spool through this and confirm it looks good I'll merge it.

plural (approved):

plural (approved):

plural (approved):

PR #749, opened 2023-02-06
Added flavour text for Qianju PT

lostgeek:
Added flavour text for Qianju PT


distributive:
You'll also need to update the [v2 version](https://github.com/NetrunnerDB/netrunner-cards-json/blob/17e71cd3fe38534730ab2ea5a23c616465bd2359/v2/printings/order_and_chaos.json#L466) of the data to pass the tests

PR #751, opened 2023-02-20
Add JSON for rulings and ruling sources.

plural:
This is based on a transformed version of the rulings data in the NRDB database now.

There is no need to review every file.  I'm using this to put a demo together and gather feedback.


NoahTheDuke:
This seems outside the scope of the json data.

plural:
if the JSON data is the source of the API's data, i don't agree,  I'd
rather have this all well integrated than living in a solo feature of a
deprecated website's database.  ;)

distributive (approved):

PR #763, opened 2023-07-23
Add translations for some NSG sets

csbisa:
This pull request adds a bunch of translated card titles for some NSG sets.

There's a few set-up commits:

- Rename jp->ja and kr->ko for existing translations: This ensure we use ISO 639-1 codes for all languages under translations.
- Rename zh to zh-hans: This sets up a split for simplified and traditional Chinese translations.
- Add skeleton directories for the remaining languages supported by NSG

Which is then followed up by adding all translated card titles for System Gateway, Midnght Sun, and Parhelion. This is the data I've recieved from NSG so far and other sets should follow in the future.

This supersedes a old pull request which only had a small amount of data (https://github.com/NetrunnerDB/netrunner-cards-json/pull/756).


plural (approved):
I'm going to go ahead and merge this even though we don't actually use this data yet.  thanks.

PR #785, opened 2023-09-01
MU notation fixes for Endurance and Masterwork.

rubenpieters:
1. Endurance is missing a space between "+2" and "MU" (in its `stripped_text`)
2. Masterwork has an errant period at the end of its "+1 MU"


plural:
Thanks!  I filed #786  to track adding tests to enforce this.

plural (approved):

PR #787, opened 2023-09-10
remove rotated cards from SBL 23.08

francescopellegrini:
So long, Data & Destiny and Flashpoint


plural:
Please fix lint check errors.

plural (approved):

PR #788, opened 2023-09-12
add SBL 23.09

francescopellegrini:
https://nullsignal.games/blog/standard-ban-list-23-09/


plural (changes requested):
Thanks for sending this.  Would you also remove the rotated cards from the new ban list entry and sync in from head?

plural (approved):

PR #791, opened 2023-11-25
Remove base_link from A Teia and Epiphany and memory_cost from runner ids.

plural:


distributive (approved):
Do we not store memory limits for Runner IDs? We probably should

plural:
> Do we not store memory limits for Runner IDs? We probably should

we don't!  I was surprised as well.  I'll file an issue for it.  We only have 1 ID that affects MU, but it's also the ID text, not a basic limit of 5 for Chaos Theory.

PR #794, opened 2023-12-16
Add 2023-11 eternal points update.

plural:


v2/formats/eternal.json
```diff
@@ -27,10 +27,17 @@
     },
     {
       "id": "eternal_4",
-      "active": true,
+      "active": false,
```
distributive, line 30: We can just omit this line.

distributive (approved):

v2/formats/eternal.json
plural: Done, thanks.

PR #797, opened 2024-01-13
Merge docker workflow to main branch to be able to test

leukosaima:
Simple docker file to start, copies contents of repo to /mnt/netrunner-cards-json which should be mounted as a volume, from which another container could read from.

New workflow to try to build and push to a container registry, unfortunately it looks like it has to be merged to main to even begin testing it.

Re purposed previous workflow to be performed only on PR's, and the new workflow will run on pushes to main, running the same tests as the previous workflow but also pushing to docker registry.


Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.19
+
+COPY . /repocontents/
```
plural, line 3: The files we need are:
./v2/
/.*.json  minus package*.json, tsconfig.json, .mocharc.json, .eslintrc.json, 
./pack/*.json
./translations/*

and technically we don't need any of the .git folders under there, just json.

plural (changes requested):

.github/workflows/docker-cicd.yml
```diff
@@ -0,0 +1,67 @@
...
+        with:
+          token: ${{ secrets.DIGITALOCEAN_ACCESS_TOKEN }}
+          
+      - name: Log in to DO Container Registry 
+        run: doctl registry login --expiry-seconds 600
+        
+      - name: Tag image 
+        run:
+          docker tag nsgcr/netrunner-cards-json \
+          registry.digitalocean.com/nsgcr/netrunner-cards-json:${{ github.run_id }}
```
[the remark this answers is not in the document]
leukosaima: sure, it's good for development.  In prod it should be avoided, i'm in the middle of handling a lot of production tasks at work and i didn't include it here out of habit

Dockerfile
leukosaima: done

plural (approved):

PR #803, opened 2024-01-20
Fix text, stripped text and flavor issues in TAI.

plural:
we had a few leftover issues with errant spaces and some actual issues (2 instead of -2) in some stripped_text.


talha-ahsan (approved):
🚀 🌔 💲 💲 💲 

Looks good to me, I'll try to figure out what's going on with accents being removed in the other cards

PR #807, opened 2024-02-06
Update System Gateway to the Remastered text.

plural:


locks (approved):
[ignore original comment]

PR #811, opened 2024-02-09
Handle flip cards in updateDuplicatePrintings.

plural:


talha-ahsan (approved):
Nobody would ever make a 3+ dimensional card right? _right_?

PR #815, opened 2024-02-24
small replacements to Makers Eye, Aesops, and Tomorrows headline

talha-ahsan:
GLC had a comment on how Tomorrow's headline in particular has the weird ' after the SU remaster update and that it's breaking JNet. 

This should revert titles to make things a bit better?

### DIFF
--- a/v2/rulings/aesops_pawnshop.json
+++ b/v2/rulings/aesops_pawnshop.json
@@ -4,6 +4,6 @@
     "card_id": "aesops_pawnshop",
     "date_update": "2022-04-23",
     "nsg_rules_team_verified": false,
-    "question": "Can [Aesop’s Pawnshop](https://netrunnerdb.com/en/card/31035) be used multiple times in a turn to trash multiple cards?"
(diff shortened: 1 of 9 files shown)


v2/rulings/aesops_pawnshop.json
```diff
@@ -4,6 +4,6 @@
     "card_id": "aesops_pawnshop",
     "date_update": "2022-04-23",
     "nsg_rules_team_verified": false,
-    "question": "Can [Aesop’s Pawnshop](https://netrunnerdb.com/en/card/31035) be used multiple times in a turn to trash multiple cards?"
+    "question": "Can [Aesop's Pawnshop](https://netrunnerdb.com/en/card/31035) be used multiple times in a turn to trash multiple cards?"
```
talha-ahsan, line 7: not sure if this one should be updated?

plural:
I am trying to fix this with a patch to NRDB instead of this.  I know people are a bit annoyed with me about this, but if you wanna do card interchange, that's what ids are for.

PR #816, opened 2024-03-02
Added pronouns for IDs

distributive:
Draft IDs weren't specified in the list I was given so they have been excluded for now


package-lock.json
plural, line 1: does this change need to be in this PR?

schema/v1/card_schema.json
plural, line 1: I don't believe you need this file.

v2/cards/khan_savvy_skiptracer.json
```diff
@@ -8,6 +8,7 @@
   "influence_limit": 12,
   "is_unique": false,
   "minimum_deck_size": 40,
+  "pronouns": "he/him",
```
plural, line 11: is this one correct?

v2/cards/apex_invasive_predator.json
```diff
@@ -8,6 +8,7 @@
   "influence_limit": 25,
   "is_unique": false,
   "minimum_deck_size": 45,
+  "pronouns": "[missing]",
```
NoahTheDuke, line 11: it/its, right?

NoahTheDuke:
I think keeping the non-NSG pronouns in is good. (Quetzal, Nero, etc.) Is it just "we don't own those" or is there another reason?

distributive:
It was a request by Narrative

plural (approved):
Nice, thanks!!!

PR #823, opened 2024-03-10
add SBL 24.03

francescopellegrini:
Not set as active yet


plural:
Thanks for being quick on the trigger, but we're handling this.

We have a new tool to generate these and I want to make sure we take that path for the new banlist.

PR #824, opened 2024-03-10
New banlist tool

plural:


talha-ahsan (approved):

src/new-banlist.ts
```diff
@@ -0,0 +1,174 @@
...
+  }
+});
+
+srcRestriction.date_start = options.effective_date;
+srcRestriction.name = options.name;
+srcRestriction.id = textToId(options.new_code);
+srcRestriction.banned.push(...addCards);
+srcRestriction.banned = srcRestriction.banned.filter((c: string) => !removeCards.includes(c));
+// TODO(plural): Remove this when prettier's jsonRecursiveSort is fixed.
+srcRestriction.banned = srcRestriction.banned.sort();
```
NoahTheDuke, line 161: `Array.prototype.sort` does the sorting in-place, so no need to re-assign it here. [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)

src/new-banlist.ts
```diff
@@ -0,0 +1,174 @@
...
+
+import fs from "fs";
+import { resolve } from "path";
+import commandLineArgs = require('command-line-args');
+import prettier from "prettier";
+import { getMwlJson, getRestrictionsV2Json, textToId, getPrintingsV2Json } from "./index";
+
+async function formatJSON(text: any) {
+  const data = JSON.stringify(text);
+  return await prettier.format(data, {
```
NoahTheDuke, line 16: You don't need to await here, as you're returning the promise either way.

src/new-banlist.ts
```diff
@@ -0,0 +1,174 @@
...
+const restrictions = getRestrictionsV2Json();
+
+const standardRestrictions = restrictions['standard'];
+
+let srcRestriction: any = null;
+
+standardRestrictions.forEach((v) => {
+  console.log(`${v.id}`);
+  if (textToId(options.source_code) == v.id) {
+    srcRestriction = v;
```
NoahTheDuke, line 151: This doesn't early exit, so it'll be overwritten/last one wins. Is that intentional?

plural:
> Given that so many functions are async and relying on async functions in top-level statements, should you make this a little more modular and put the bulk of the work into functions that are called one after the other? That'll let you say `await formatJson(); await processArgs(); await applyChanges();` etc.

yeah, i probably should.  i don't want any of this to be async, but pesky JS lib authors love that stuff.

src/new-banlist.ts
```diff
@@ -0,0 +1,174 @@
...
+const printings = getPrintingsV2Json();
+printings.forEach(p => {
+  if (!cardsToPrintings.has(p.card_id)) {
+    cardsToPrintings.set(p.card_id, new Array<string>());
+  }
+  cardsToPrintings.get(p.card_id)?.push(p.id);
+});
+let addCards = new Array<string>();
+if (options.add_cards) {
+  addCards = options.add_cards.split(',');
```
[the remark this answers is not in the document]
plural: these will be card ids, not names, so they don't have commas.

src/new-banlist.ts
```diff
@@ -0,0 +1,174 @@
+// Create a new banlist based on an existing banlist and additions/deletions.
+//
+// Currently only handles standard implicitly.
+// Example usage:
+//   npm run build
+//   npx ts-node dist/new-banlist.js -s standard-ban-list-23-09 -n "Standard Banlist 24.03" --effective_date="2024-03-18" -c standard-ban-list-24-03 -a formicary
+
+import fs from "fs";
+import { resolve } from "path";
+import commandLineArgs = require('command-line-args');
```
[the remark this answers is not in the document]
plural: yep.

src/new-banlist.ts
plural: ah, thanks.

src/new-banlist.ts
plural: an early exit isn't needed.  The repo tests enforce that there aren't dupes in the JSON.

plural:
I tidied this up a bit.  I appreciate the review, Noah.

PR #827, opened 2024-03-13
Rebellion Without Rehearsal spoilers day 6!!!

plural:


NoahTheDuke:
i appreciate your exclamation points, they're quite fun

PR #829, opened 2024-03-15
Rebellion Without Rehearsal Spoilers Day 8!!!!!!!!

plural:


talha-ahsan (approved):
I just realized you're scaling up the number of exclamation points.

PR #834, opened 2024-03-17
Change start date for Sunset to move it back in time.

plural:
This pushes sunset down in the list.  We can't delete it without work that probably isn't worth doing.

### DIFF
--- a/mwl.json
+++ b/mwl.json
@@ -1491,7 +1491,7 @@
       "33111": { "deck_limit": 0 }
     },
     "code": "sunset-ban-list-24-01",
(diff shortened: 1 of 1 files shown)


NoahTheDuke:
Can't set it to the real date?

plural:
> Can't set it to the real date?

no because NRDB Classic would have made it the active standard ban list if we had done that. Since the team wants this to be done, at least this puts it a bit farther down the list.  no date is good with this hack in classic.  :/

talha-ahsan (approved):

PR #840, opened 2024-05-03
TAI pronunciation guide

talha-ahsan:
RwR is still having some changes so I'm waiting for a thumbs up before porting those over

### DIFF
--- a/v2/cards/bahia_bands.json
+++ b/v2/cards/bahia_bands.json
@@ -7,6 +7,8 @@
   "id": "bahia_bands",
   "influence_cost": 1,
   "is_unique": false,
+  "pronunciation_approx": "BYE-yah",
+  "pronunciation_ipa": "bɐˈiɐ",
   "side_id": "runner",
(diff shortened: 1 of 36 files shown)


v2/cards/bahia_bands.json
```diff
@@ -7,6 +7,8 @@
   "id": "bahia_bands",
   "influence_cost": 1,
   "is_unique": false,
+  "pronunciation_approx": "BYE-yah",
```
plural, line 10: any reason this doesn't have bands in here as well?

plural: a similar question for other cards.  Why does the pronunciation only have part of the title?

If that's the convention, cool cool, but i'm curious.

plural (approved):

v2/cards/bahia_bands.json
talha-ahsan: This is more or less the convention AFAIK. I'm assuming English words with known pronunciation by native English speakers were omitted. They're not in the guide, and I don't trust myself to generate proper IPA for them on my own

plural: that's what I expected.  Thanks for confirming.  :)

plural (approved):

PR #844, opened 2024-07-01
Fix unclosed em tags in flavor text.

### LINKED ISSUE #836: Correct Mantle Flavor Text Emphasis Tags
If you visit the NRDB page for [Mantle](https://netrunnerdb.com/en/card/26088), you can see lots of italics leaking around the page in unexpected spots as compared to some other card like the System Gateway printing of [Sure Gamble](https://netrunnerdb.com/en/card/30030).

The culprit appears to be the flavor text. It contains "em" tags, but it contains two opening "em" tags instead of an opening "em" tag and closing "/em" tag.

This is at line 451 in ur.json.

For what it's worth, while researching this, I noticed Sure Gamble uses the "strong" tag, which are correctly opened and closed.  However, it makes the text on NRDB bold instead of the un-italicized text seen on the physical card.

plural:
Fixes #836  and the same issue for Stargate.

### DIFF
--- a/pack/df.json
+++ b/pack/df.json
@@ -65,7 +65,7 @@
     "deck_limit": 3,
     "faction_code": "anarch",
     "faction_cost": 3,
-    "flavor": "“Net <em>space<em> is an abstraction, a white lie protecting fragile comprehensions. Do not limit yourself.”\n)—z\\h/r",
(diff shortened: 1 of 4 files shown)


talha-ahsan (approved):
Does the import tool also need an update for the `<em>` rule for this?

plural:
i'll add a test to the JSON repo for unclosed tags.

PR #849, opened 2024-09-07
Update card pools for most recent standard banlists.

plural:
Fixes #840

### DIFF
--- a/v2/formats/standard.json
+++ b/v2/formats/standard.json
@@ -152,7 +152,7 @@
       "restriction_id": "standard_ban_list_23_08"
     },
     {
-      "active": true,
+      "active": false,
       "card_pool_id": "rotation_2023",
       "date_start": "2023-09-22",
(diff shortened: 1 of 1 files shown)


v2/formats/standard.json
```diff
@@ -165,6 +165,20 @@
...
     },
+    {
+      "active": true,
+      "card_pool_id": "rotation_2023",
+      "date_start": "2024-05-31",
+      "id": "standard_27",
+      "restriction_id": "standard_ban_list_24_05"
+    },
+    {
+      "active": false,
```
talha-ahsan, line 176: shouldn't 24_09 be active and 24_05 be not active?

plural: I meant to respond in here - The new banlist isn't effective until Friday, so this is technically correct until then.

talha-ahsan (approved):

PR #858, opened 2024-12-13
24.12 banlist

talha-ahsan:
This also removes some rotated currents from the ban list


v2/restrictions/standard/standard_ban_list_24_12.json
```diff
@@ -0,0 +1,58 @@
...
+    "ssl_endorsement",
+    "system_seizure",
+    "tithonium",
+    "tributary",
+    "trick_shot",
+    "watch_the_world_burn",
+    "whampoa_reclamation",
+    "world_tree"
+  ],
+  "date_start": "2024-12-12",
```
plural, line 53: should be  2024-12-26

plural (changes requested):

v2/restrictions/standard/standard_ban_list_24_12.json
talha-ahsan: fixed

plural (approved):

PR #859, opened 2024-12-14
Add banlist 24-12 to standard format.

plural:
needed to tie this to the format.  Without this it doesn't show up on the NRDB classic ban list page.


plural:
Going to get this and a test suite update in without review to get NRDB Classic all up and running.

PR #875, opened 2025-04-01
Astroscript un-errata, new points

talha-ahsan:
This does mean we need the update the image for Astroscript to not have the L1PD


plural (approved):

plural:
https://netrunnerdb.com/card_image/01081_original.png is the source for the card image now.  We will need to do an updated image version folder for this.

plural (approved):

PR #907, opened 2025-06-04
Public Access Plaza has no subtype.

plural:


NoahTheDuke:
why are you nerfing Spark

plural:
> why are you nerfing Spark

lol

PR #912, opened 2025-08-10
fixed whiteblades name

talha-ahsan:


NoahTheDuke:
sam, not samantha?

talha-ahsan:
Yep, I explicitly asked if they wanted to be "Sam" or something else, so for now it's just Sam

plural (approved):

PR #916, opened 2025-09-17
29 10 sbl

talha-ahsan:


v2/formats/standard.json
```diff
@@ -202,6 +202,12 @@
       "date_start": "2024-08-01",
       "id": "standard_31",
       "restriction_id": "standard_ban_list_25_08"
+    },
+    {
+      "card_pool_id": "rotation_2025",
+      "date_start": "2024-10-03",
+      "id": "standard_31",
```
migueldlr, line 209: Should this ID be incremented?

PR #976, opened 2026-05-18
Ignore ds store files

plural:


locks:
Can you rebase @plural?

locks (approved):

PR #977, opened 2026-05-30
Update and fix for french translations

galehar:
I have added the missing french translations for Ashes, Liberation and Elevation. I also fixed a few incorrect ones in System Gateway and Borealis


galehar:
oh, I forgot to mention. I know nrdb currently doesn't support  i18n, but jnet does and is pulling translations from here. See [the related jnet issue](https://github.com/mtgred/netrunner/issues/8744)

plural (approved):

PR #985, opened 2026-07-19
V2 translation prep

plural:


src/update-locales-v2.ts
migueldlr, line 1: This file is pretty large, is there any way we could break this down into more composable pieces?

migueldlr (approved):

src/update-locales-v2.ts
plural: i pulled the IO out which brought it down a bit more.  I'd like to keep this as-is for now and include this as part of some broader refactoring followups.

PR #992, opened 2026-08-10
Add core format and card pool

migueldlr:


plural (approved):
hell yeah.  Thanks for adding this.

